PR #93, opened 2014-07-21
Change to `use_2to3=False` in `setup.py`

cdeil:
This PR changes to `use_2to3 = False` for `package-template`, which I think should be the default.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,6 +45,9 @@ distribute-*.tar.gz
 # Other
 .*.swp
 *~
+.project
+.pydevproject
+.settings
 
 # Mac OSX
 .DS_Store
--- a/packagename/__init__.py
+++ b/packagename/__init__.py
@@ -12,4 +12,4 @@ from ._astropy_init import *
 
 # For egg_info test builds to pass, put package imports here.
 if not _ASTROPY_SETUP_:
-    from example_mod import *
+    from .example_mod import *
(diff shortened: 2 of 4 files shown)


cdeil:
By now there's only one official affiliated package, `pyvo`, that still uses `use_2to3=True`.

And what just happened was that @keflavich accidentally changed `astroquery` back to `use_2to3=True` when merging in updates from `package-template` (see https://github.com/astropy/astroquery/pull/485).

So I don't think the argument "changing the default in package-template will break existing affiliated packages" holds any more and still think setting `use_2to3=False` in the package-template would be better.

pllim:
:+1: to merging this.

saimn:
:+1: to merge this. New affiliated packages should not be encouraged to use 2to3. And there are probably very packages that actually use it.

cdeil:
By now using `2to3=True` in `setup.py` really isn't considered best practice anymore (see e.g. the official https://docs.python.org/3/howto/pyporting.html ), so I don't think we should advertise a package template that does this.

@astrofrog - You said no to `2to3=False` in default `setup.py` for `package-template` over two years ago. Time to re-consider this PR now?

astrofrog:
Let's do it! :)

cdeil:
😄 
```python
from __future__ import thank_you
```

PR #98, opened 2014-07-24
Modified README.rst to accommodate users starting from an existing repo

aphearin:
I made some relatively minor changes to README.rst in the package template. The current version on astropy is written for users who are starting entirely from scratch. Experienced git users probably wouldn't have much trouble figuring out how to adapt the current astropy README to preserve their own git history while still being able to occasionally pull in new changes to the package-template. But this is a non-trivial task for newer git users, and the modifications to the instructions I made are intended to address that subset of people. 

@eteq


astrofrog:
This is no longer relevant for this repository (since the instructions have been moved to the core package)

PR #113, opened 2015-01-05
Include Cython .pxd files in setup.py

larrybradley:
This PR adds Cython `.pxd` files (like C header files) to be included in the distribution.  When present, the `.pxd` files are needed for compilation.  I first encountered missing `.pxd` files when packaging the `photutils 0.1` release.


larrybradley:
`travis-ci` passed now with #114.

setup.py
```diff
@@ -84,7 +84,7 @@
 c_files = []
 for root, dirs, files in os.walk(PACKAGENAME):
     for filename in files:
-        if filename.endswith('.c'):
+        if filename.endswith('.c') or filename.endswith('.pxd'):
```
mdboom: Minor point, but should we rename the `c_files` variable while we are at it (since it's no longer accurate)?  Maybe "compiled_files"?

larrybradley: Good idea, thanks @mdboom.  I made that change.

embray:
You know, I don't page a lot of attention to the package-template, so I hadn't really scrutinized this code before, but what it's doing is not correct (in the first place, not so much your modification to it, @larrybradley).  The [`package_data`](https://docs.python.org/2/distutils/setupscript.html#installing-package-data) argument to `setup()` is specifically for files that should be *installed* alongside Python packages.  Although any files listed in `package_data` will be included in the source distribution, it is *not* for the purpose of specifying files to distribute in general.

What this results in is .c files being installed as part of the package, and that's not good.  This code should be removed, and a different solution sought for whatever problem it was trying to solve (this could include something in astropy-helpers if need be).

larrybradley:
Thanks, @embray.  I'm not exactly sure what the general problem is/was.  It looks like @mdboom wrote the original solution, so I'll leave it to him.  :)

mdboom:
Yep.  The solution it's trying to solve is to include these files in the source distribution.  I couldn't find a better way...

embray:
The use of MANIFEST.in for this (as opposed to any other means of listing files) is an unfortunate historical kludge, but I think it's the way to go, and affiliated package authors should simply know how to edit this file.

Which is not to say I'm opposed to providing better ways either.  setuptools does have a plugin interface that allows defining additional file_finders that can implement whatever rules we want for what files should be include in a source distribution (this is how various VCS plugins are supported, but I don't think we necessarily want as a default to include all files under version control in the source dist).  

Perhaps I could write something up to include in astropy-helpers that makes it easier to do this without having to edit MANIFEST.in directly.

cdeil:
@larrybradley – Can you please rebase this PR on master?

larrybradley:
@cdeil Rebased.

embray:
Sorry, this really shouldn't be merged though.  And the existing code that adds the .c files to package_data needs to be removed as well.  I'll open a separate change in astropy-helpers to make it easier to include these files via the sdist command.  In the meantime you should just edit MANIFEST.in.

PR #119, opened 2015-02-24
Changed the import from astropy to astropy_helpers 

crawfordsm:
In the conf.py in the docs directory, it was trying to import from astropy and not astropy_helpers if the edit_on_github parameter was set.   This is a small change to switch to the correct import statement so it compiles and build the documentation.  

The error that was being produced was:
```
Traceback (most recent call last):
  File "<stdin>", line 27, in <module>
  File "/usr/local/lib/python2.7/site-packages/sphinx/application.py", line 119, in __init__
    self.setup_extension(extension)
  File "/usr/local/lib/python2.7/site-packages/sphinx/application.py", line 336, in setup_extension
    err)
sphinx.errors.ExtensionError: Could not import extension astropy.sphinx.ext.edit_on_github (exception: No module named edit_on_github)

```

Related to https://github.com/astropy/astroquery/issues/478

PR #120, opened 2015-02-26
Update conf.py to use astropy_helpers.sphinx

eteq:
This updates the conf.py file in the docs to use `astropy_helpers.sphinx` instead of `astropy.sphinx` (which was removed in v1.0).

@astrofrog @embray - I couldn't find anywhere obvious in the changelogs indicating that `astropy.sphin` was removed in v1.0... It was, wasn't it?  Is it documented somewhere?


eteq:
Whoops!  I just noticed #119 where @crawfordsm did the exact same thing.  Will close this in favor of that.

PR #122, opened 2015-03-26
Monkey patch for config bug in astropy 1.0.1

mdboom:
This makes sure that the config system works for affiliated packages, even on 1.0.1 where certain things related to automatic introspection were broken.


mdboom:
It seems only astroquery is affected by this bug, and a version of this is already merged there.  Closing, because I don't see a need to muddy other affiliated packages with this hack unnecessarily.

PR #128, opened 2015-05-28
Remove requires and provides from setup.py

astrofrog:
As suggested by @embray in https://groups.google.com/forum/#!topic/astropy-dev/urYO8ckB2uM


embray:
:+1:

embray:
It's kind a shame too since it does still show that info on PyPI and is reasonably useful in practice.  But at the end of the day it's not worth the confusion over these keywords.

astrofrog:
@embray - should we actually add a comment to explain why they are missing in case anyone thinks about adding them back?

embray:
If you want.  You could just link to my post where I explained it?  Or be terse and claim "they're deprecated"

PR #129, opened 2015-07-15
Added version to setup.cfg.

cmccully:
I have added the version number to setup.cfg in hopes of making it more intuitive to find.

PR #130, opened 2015-07-15
Moved entry points to setup.cfg.

cmccully:
I moved the console entry scripts into setup.cfg.


embray:
:+1: 

For what it's worth, at some point I plan on moving most of astropy_helpers (and hence astropy and the affiliated package template) over to using d2to1 (or its successor that I'm thinking about): https://pypi.python.org/pypi/d2to1

Changes like this one and #129 will help with that transition.

PR #131, opened 2015-07-15
Added install requirements to setup.cfg.

cmccully:
I believe that this pull request with #129, #130, #93, the affiliated packages should only need to edit setup.cfg but not setup.py.


cmccully:
I am definitely open to discussion on the best way to do this.

embray:
:-1: on requirements.txt.  If I recall correctly, I'm not sure if pip even reads this to install runtime requirements for a package by default.  Rather, the general distinction between a requirements.txt and install_requires is that install_requires tends to be as loose as possible with version numbers, while requirements.txt is usually used to pin requirements, for example to reproduce a development environment.

This is just a general guideline though, and is at least what I last heard on the issue last it was discussed on distutils-sig.  Instead I'd be in favor of putting install-requires in the setup.cfg, as is supported by d2to1, and leave out a separate requirements file.

cmccully:
@embray Fair enough. I would prefer the setup.cfg file as well. I have updated it. What do you think?

embray:
Looks great, thanks!

embray:
I would go ahead and merge, but I think @astrofrog or @eteq might want a look too.  Actually I'm not sure anymore who should be tagged for changes in package-template.  We should have a list somewhere of "package-template stakeholders" ... :)

astrofrog:
:+1:

embray:
Before merging this should we make some announcements about package-template changes?  This could (mildly) break some packages if their maintainers blindly merge it without also updating their setup.cfg.

mwcraig:
I think that would  be helpful...perhaps with some advice as to whether current affiliated packages should move in this direction?

astrofrog:
@embray - I would say that once all these changes are merged in, we tag a new version and make an announcement then?

We should add an entry to ``TEMPLATE_CHANGES.md`` too

embray:
This needs to be rebased now since I merged #130, and maybe also an entry added to TEMPLATE_CHANGES.md (for #130 as well, now that I think about it) as @astrofrog suggested.

cmccully:
I just rebased this pull request. Sorry for the delay. Is there anything else that needs to be done before this can be merged, @embray?

embray:
I think mistakes were mad in the rebase :( https://github.com/astropy/package-template/pull/131/files

cmccully:
@embray You were correct that I missed one of the merge comments. Sorry about that. I fixed the typo and pushed the fix. I also noticed a couple more minor issues that I fixed. The only significant change I made was to switch from using install_requires to setup_requires. I noticed that if numpy wasn't installed previously and was only included in install_requires, the setup would crash. I am still waiting for the travis build to finish but this builds correctly on a clean environment on my local machine.

cmccully:
Also, should sphinx be a setup requirement?

cmccully:
@embray  @astrofrog  @eteq The only test that is failing is the test --coverage command. This has been noted in other astropy repos. Should we pin the version of coverage? Or realistically should we update the base package that uses the coverage module?

setup.py
```diff
@@ -98,15 +98,20 @@
...
 # Note that requires and provides should not be included in the call to
 # ``setup``, since these are now deprecated. See this link for more details:
 # https://groups.google.com/forum/#!topic/astropy-dev/urYO8ckB2uM
-
 setup(name=PACKAGENAME,
       version=VERSION,
       description=DESCRIPTION,
       scripts=scripts,
-      install_requires=['astropy'],
+      setup_requires=setup_requires,
```
embray: I don't really understand what's going on here anymore.  `setup_requires` and `install_requires` are totally different things, nor should we be imposing the above defaults.

cmccully: The above defaults are meant to stop failures from using the astropy_helpers module which is included in the package template. If any of those 3 packages are not installed, running python setup.py will crash. By adding those packages to setup_requires. If those packages are not installed, distutils will try to automatically install them. I think it might make sense to include these defaults in setup_requires and take the install_requires from the setup.cfg file. What do you think?

embray: That isn't true, or at least, it's not supposed to be. If it is, that's another problem.  And in any case requires-dist is meant to be used for runtime requirements, not build-time requirements.  The way `setup_requires` and `install_requires` work are very different.

cmccully: I think this issue is unique to packages that rely on Cython (e.g. astroscrappy). If cython is not included in setup_requires, then the build will obviously not work because it can not generate the c source files. To get the numpy linking correct, often numpy.get_include() is used, which also requires numpy to be in setup_requires.
Upon further inspection, I believe that astropy does not need to be in setup_requires. 

My thought would be to do something like:
setup_requires = ['numpy', 'cython']
install_requires = install_requires
where install_requires is from the setup.cfg file. What do you think? Is there a cleaner way to fix the cython issue?

embray:
I'll comment more later, but this is also related:
astropy/astropy#3229
astropy/astropy-helpers#185
astropy/astropy-helpers#167

astrofrog:
Note that the current implementation is incorrect as-is in this pull request, because running

```
python setup.py egg_info
```

now installs all the dependencies from source, which is undesirable.

astrofrog:
@cmccully - replying to your comment in astropy/astropyscrappy#2 here since it applies here as well. Basically, you need to do what astropy does:

```
# Avoid installing setup_requires dependencies if the user just
# queries for information
if is_distutils_display_option():
    setup_requires = []
```

See https://github.com/astropy/astropy/blob/master/setup.py#L84

Basically ``setup_requires`` should always be empty if ``is_distutils_display_option()`` is True.

cmccully:
I have updated this PR to account for @astrofrog's comment so now nothing is installed for egg_info. I have also updated things so that I think everything makes sense in terms of what needs to be in setup_requires vs install_requires per discussion with @embray. Is there anything else that needs to be done before this can be merged?

setup.py
keflavich: Once this is resolved, could one of you include a 'best practices' recommendation in the package-template for how and when `setup_requires` and `install_requires` should be populated?  e.g., what to do with (1) packages required at runtime, but not for install (2) packages required at runtime (3) optional packages that users might want to install but are not strictly required

cmccully:
@astrofrog @embray @eteq Is there anything else that needs to be done to this PR before it can be merged?

eteq:
@cmccully - I'm happy with this *except* for the lingering `install_requires`/`setup_requires` question.  It looks to me like we're still in a situation where *anyone* using the template will end up with `numpy` and `cython` as setup requirements, and a lot of affiliated packages require neither at setup time (basically anything that's pure python). This is particularly problematic because (at least, we think), `pip` doesn't always play nicely with `setup_requires`.

So here's a possible solution: let the user decide.  I see two possible approaches to this:
1. read the requirements from ``setup.cfg`` and if the user specifically mentions Cython or numpy, interpret that as it needing to be in `setup_requires`
2. add another ``setup.cfg`` entry that could be just something like ``needs_c_compiling = yes/no``.

Also, is there a companion PR to update http://astropy.readthedocs.org/en/latest/development/affiliated-packages.html ?  If I understand right, after this there's no need to edit ``setup.py`` (a good thing!) but then the instructions should be updated to reflect that.

cmccully:
Thanks @eteq. I think I prefer option 1 that you suggest. That seems like a very sensible compromise. I will update the PR implementing this idea. 

I'm surprised to hear that pip doesn't play nicely with setup_requires. I use setup_requires with astropy/astroscrappy and it seems to work without issue. Could you give me more details about the failures?

There is not a corresponding PR for the docs yet. I was hoping to get this merged along with #129 and do one PR on the documentation. I will get that started.

eteq:
I *think* it's just that ``pip`` options that affect versioning and such only get applied to regular dependencies, not ``setup_requires`` (the specific case I've seen is when asking pip not to do upgrades, and it ignores that for ``setup_requires``).  So it does install things that are in ``setup_requires``, but sometimes behaves in ways that users consider unexpected when doing anything other than a standard ``pip install <something>``.  Hence we're in a situation of "usually it works, but occasionally not as expected".

You could maybe make an argument that this is intended behavior: ``setup_requires`` could be interpreted as "must be exactly like this or setup won't work".  I don't honestly know for sure if it's actually intended or just an oversight, though.

embray:
I'm thinking of actually vendoring a copy of d2to1 into astropy-helpers, which would make this a moot point.

cmccully:
@eteq @embray Sorry about the long delay on this. I just updated the package template to account for your comments. Is this an acceptable solution?

cmccully:
@eteq @embray Any word on this?

astrofrog:
@cmccully - sorry for dropping the ball on this. Can you rebase this and also undo the changes to astropy-helpers?

setup.py
astrofrog: >  often numpy.get_include() is used

@cmccully - ah, but I believe that the way astropy-helpers works, if you are setting up Cython extensions you can just use ``'numpy'`` instead of ``numpy.get_include()`` and astropy-helpers does the substitution later. So I think ``setup_requires`` isn't actually needed.

astrofrog:
I've now implemented this much more simply in https://github.com/astropy/package-template/pull/208 for now - and left any controversial parts (in particular setup_requires) out for now.

astrofrog:
For setup_requires, I think a better solution would be to simply check if extensions are present in setup.py and adjust setup_requires accordingly, but not list setup_requires in setup.cfg. I'll work on a PR for that to see if it works, and we can discuss this further in #209.

PR #133, opened 2015-07-20
Switch to container-based Travis infrastructure

astrofrog:
cc @bsipocz @mwcraig since you did this in other packages

@eteq @embray - any objections to this? I think most packages will be fine with using this and don't need sudo?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,18 @@ python:
     - 3.3
     - 3.4
     # This is just for "egg_info".  All other builds are explicitly given in the matrix
+
+# Setting sudo to false opts in to Travis-CI container-based builds.
+sudo: false
+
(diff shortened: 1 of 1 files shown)


mwcraig:
I think this makes sense. Maybe add a link to the white list of apt packages in the README or something: https://github.com/travis-ci/apt-package-whitelist/blob/master/ubuntu-precise

mwcraig:
Was also thinking an email to astropy-dev about how to upgrade would be helpful -- I'll send out in a few hours, unless you want to do it before then.

astrofrog:
@mwcraig - thanks, added the URL

mwcraig:
Hearing no objections, I'm going to merge this...

PR #137, opened 2015-09-29
Remove .c files from package_data

embray:
Reminded by #136, I realized that this bit of code is still in the setup.py.  This should not be here and we should not be perpetuating this bad behavior.

Putting files in package_data means that those files get _installed_ when installing the Python code.  We should not be installing C source files.  

The problem that this code was solving, in part, is that the "MANIFEST.in" in package-template was incorrect (#136 attempts to fix this).  I think somebody told me at one point that they didn't want people using the package template to have to edit MANIFEST.in, but I think they'll have to for now since, for better or worse (worse, mostly) that's the correct way to control what non-installed files get into a Python source distribution, and that should just be part of the instructions.

There is an issue astropy/astropy-helpers#53 to try to automate things like this better, but in the meantime we shouldn't be promoting bad practice.


astrofrog:
This makes sense. Note that the coverage build is now fixed in master, so you can ignore that failure.

embray:
The problem is that before I merge this documentation changes will need to be made, and people using this package template will need to be alerted to the changes.

astrofrog:
@embray - can you rebase this to make sure the tests pass? Please also include a note in ``TEMPLATE_CHANGES.md`` that will then later be communicated to package maintainers.

embray:
I can try that.  It would be very nice to have this fixed.

astrofrog:
This is no longer needed with #375

PR #141, opened 2015-11-23
Assumptions about data subpackage and package_data

adrn:
In one of my projects based off of this package template, I have a `data` subpackage for loading and interacting with data -- not a container for data files. The glob pattern `data/*` in `setup.py` was picking up sub-directories and trying to add them to package data, causing errors like:

```
error: can't copy 'ophiuchus/data/__pycache__': doesn't exist or not a regular file
```

In this PR, I've added support for a `package_data_path` keyword in `setup.cfg`, which can be set to `None` to turn off the default assumption. Not sure this is the best solution, but it's an idea...

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -76,7 +76,7 @@ package_info = get_package_info()
 
 # Add the project-global data
 package_info['package_data'].setdefault(PACKAGENAME, [])
-package_info['package_data'][PACKAGENAME].append('data/*')
+package_info['package_data'][PACKAGENAME].append('data/*.*')
 
 # Define entry points for command-line scripts
 entry_points = {'console_scripts': []}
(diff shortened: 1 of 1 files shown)


embray:
I don't want to add any more of these setup.cfg "metadata" options for now (or we do just call it "package_data").  But for your own project you could certainly remove that default (and as this demonstrates I'm not sure it's a good default to have anyways, since it will break any time someone adds a `.data` subpackage to their package, and it's not *necessarily* always going to be the name someone wants to use for package-relative resource files.

embray:
Although even calling it "package_data", I wouldn't want it to go into the "metadata" section since it's not really metadata about the package.

Part of the reasoning behind what I'm saying is that I'm planning to move builds over to either [`d2to1`](https://pypi.python.org/pypi/d2to1) or something very similar to it (possibly even incorporating its functionality into astropy-helpers), and this brings with it an existing format for the setup.cfg file.  I don't want to go adding stuff that contradicts that format, such that people will have to change things again when we switch over to that.

adrn:
Sounds good to me! Given that, I'm fine closing this. In my own project I just commented out that line in  `setup.py`. Thanks!

PR #144, opened 2015-12-03
Remove adjust_compiler

eteq:
This updates the helpers to the latest master, which includes a change that makes the ``adjust_compiler`` call unnecessary by building it into `build_ext` (astropy/astropy-helpers#76).

Note that astropy/astropy#4369 does the same thing in the core

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,7 +16,7 @@ else:
 builtins._ASTROPY_SETUP_ = True
 
 from astropy_helpers.setup_helpers import (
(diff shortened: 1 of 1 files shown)


embray:
Ah okay, here you *did* update the astropy_helpers submodule.  I would just do the same on the astropy PR :)

pllim:
Can we rebase and merge this? I was attempting to use the template and ran across the deprecation warning related to this.

pllim:
Never mind. I see that this is a duplicate of #154.

PR #150, opened 2016-02-03
Remove no-longer-needed adjust_compiler

larrybradley:


larrybradley:
Sorry, just saw that is implemented in PR #144 (unmerged).

PR #158, opened 2016-03-30
Added subpackage data setup example

pllim:
It is not very obvious on what to do when subpackage itself has its own package data. So, I added an example here.

I also added tests directories for all subpackages, to illustrate that user can test each subpackage independently.


pllim:
@astrofrog , :+1: for this one but I am afraid I am *a little* biased.

PR #163, opened 2016-04-21
Added placeholder for custom project logo

pllim:
Added placeholder for users to add their own project logos to upper left corner on sidebar. This incantation is usually included in standard `conf.py` from `sphinx-quickstart` but somehow missing in the template.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -107,6 +107,10 @@ release = package.__version__
 # Custom sidebar templates, maps document names to template names.
 #html_sidebars = {}
 
+# The name of an image file (relative to this directory) to place at the top
+# of the sidebar.
+#html_logo = ''
+
 # The name of an image file (within the static path) to use as favicon of the
 # docs.  This file should be a Windows icon file (.ico) being 16x16 or 32x32
 # pixels large.


bsipocz:
Hmm, this is weird. None of the packages that have custom logo (e.g. photutils, ccdproc) use the logo this way. Have you tried to build do docs with it? How does it look compared to the usual astropy style?

pllim:
It would look something like this (the logo is set to 200px wide):

screenshot

This is to be somewhat consistent with some other documentations we have from packages that do not use Astropy template. For instance, see [HST/WFC3 documentation](http://ssb.stsci.edu/doc_tmp/stsci_python_x/wfc3tools.doc/html/index.html) by @sosey.

astrofrog:
@pllim - as a side note, there's actually no reason you can't use the exact same theme as for other projects - there is no obligation to use the astropy bootstrap theme?

pllim:
@astrofrog , true. But for now, using different themes make it obvious (for me, at least) which of our packages use Astropy template and which ones do not. (Also, I am lazy to change what already works.)

bsipocz:
@pllim - OK, this actually looks nice. However you still have the top bar saying astropy, so it's super obvious you're using the template while wfc3tools doesn't ;)

pllim:
@bsipocz , does that top bar not go away if I switch to a different theme or use custom CSS? I am not very familiar with the various behaviors of Sphinx theme layouts.

bsipocz:
Yes, I think it should go away by changing the theme.

bsipocz:
This is actually a nice and easy way to customize, so merging now.

PR #183, opened 2016-07-05
Removed Python 2.6 in travis.yml.

cmccully:
Since Python 2.6 is no longer supported by astropy 1.2, the travis tests fail for python =2.6. We should either freeze the astropy version to 1.1 here or we should remove python 2.6 from the default travis matrix. I have opted for the latter as that seems the most consistent with the astropy core development.


pllim:
I removed 2.6 for Appveyor in #170 but don't know why I didn't do that for Travis as well.

PR #281, opened 2017-10-25
fix install_requires bug

Cadair:


Cadair:
This makes install requires not a question we ask the users. Any objections to this?

astrofrog:
Why not allow users to set that?

Cadair:
it's still changeable once the template is rendered, it's just not a question asked during the "wizard" bit on running "cookiecutter".
My main reason for this is how many people when initially templating a package know what their install deps are?

Cadair:
bump

Issue #345, opened 2018-06-21
Warnings generated by setup.py with new cookiecutter package


dkirkby:
I just followed [these instructions](http://docs.astropy.org/projects/package-template/en/latest/), which went smoothly, but I noticed the following warning (2x) the first time I ran `python setup.py ...`:
```
/Users/david/Cosmo/DESI/code/skysim/astropy_helpers/astropy_helpers/setup_helpers.py:361: 
AstropyDeprecationWarning: Use of the exclude parameter is no longer supported since it does not 
work as expected. Use add_exclude_packages instead. Note that it must be called prior to any other
calls from setup helpers.
  AstropyDeprecationWarning)
```
Also, this warning appears each time I run `python setup.py ...` (I used the default initial version string):
```
/Users/david/anaconda3/envs/desi/lib/python3.6/site-packages/setuptools/dist.py:397:
UserWarning: Normalizing '0.0.dev02' to '0.0.dev2'
  normalized_version,
```

pllim:
For the first one, it was fixed in astropy/astropy-helpers#379

dkirkby:
The normalization rule triggering the `UserWarning` is specified in [PEP440](https://www.python.org/dev/peps/pep-0440/#integer-normalization), and removes leading zeros.

However, the cookiecutter default version is not causing this:
```
"project_version": "0.0.dev0",
```
I guess the problem is somewhere in astropy-helpers, maybe [get_git_devstr](https://github.com/astropy/astropy-helpers/blob/master/astropy_helpers/git_helpers.py#L61)?

dkirkby:
Digging deeper, the version problem is in this package after all.  The cause is [this line](https://github.com/astropy/package-template/blob/cookiecutter/%7B%7B%20cookiecutter.package_name%20%7D%7D/setup.py#L91) in setup.py:
```
VERSION += get_git_devstr(False)
```
The simplest fix is to replace `dev0` with `dev` in three places.  A more robust solution would also detect any trailing integer after `dev` and generate a warning/error or silently remove it before appending `get_git_devstr()`.

dkirkby:
Related to this, the PEP386 reference in `setup.cfg` could be updated to PEP440.

PR #346, opened 2018-06-21
Use latest astropy-helpers

dkirkby:
This is to fix the deprecation warnings mentioned in #345.


astrofrog:
Thanks! I just merged https://github.com/astropy/package-template/pull/339 which did this but also removed the ``ez_setup.py``, so will close this.

PR #355, opened 2018-08-15
Allow render commit to be empty

astrofrog:
Sometimes, changes to the template don't result in changes to the rendered version but if we don't allow empty commits, the render script crashes on the empty commit.


astrofrog:
I'm just going to merge this since no point in running the CI (this change won't run)

PR #371, opened 2018-10-26
Make setup.py valid on Python 2 always for error printing reasons

Cadair:
If you run setup.py on a package which doesn't support python 2 it should print the error saying it supports Python 3.x+ or whatever. For this to work we need to keep the setup.py file syntax compatible with all python versions for as long as we want this error to work.

Note: the file only has to be runtime compatible with Python 2 above the version check error!

### DIFF
--- a/{{ cookiecutter.package_name }}/setup.py
+++ b/{{ cookiecutter.package_name }}/setup.py
@@ -1,5 +1,7 @@
 #!/usr/bin/env python
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
+# Note: This file needs to be Python 2 / <3.6 compatible, so that the nice
+# "This package only supports Python 3.x+" error prints without syntax errors etc.
 
 import glob
 import os
--- a/{{ cookiecutter.package_name }}/setup.py
+++ b/{{ cookiecutter.package_name }}/setup.py
@@ -4,14 +4,10 @@
 import glob
 import os
 import sys
-{% if cookiecutter.minimum_python_version.startswith("2") %}
 try:
     from configparser import ConfigParser
 except ImportError:
     from ConfigParser import ConfigParser
-{%- else %}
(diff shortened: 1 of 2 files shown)


{{ cookiecutter.package_name }}/setup.py
```diff
@@ -4,14 +4,10 @@
 import glob
 import os
 import sys
-{% if cookiecutter.minimum_python_version.startswith("2") %}
 try:
     from configparser import ConfigParser
 except ImportError:
     from ConfigParser import ConfigParser
-{%- else %}
```
astrofrog: Probably should add a comment about this otherwise someone will add this back again in future

Cadair: damn it, thought I did, must have undone it by mistake.

{{ cookiecutter.package_name }}/setup.py
```diff
@@ -1,17 +1,15 @@
 #!/usr/bin/env python
 # Licensed under a 3-clause BSD style license - see LICENSE.rst
+# Note: This file needs to be Python 2 / <3.6 compatible, so that the nice
+# "This package only supports Python 3.x+" error prints without syntax errors etc.
```
astrofrog: Maybe put it around the configparser import since that's where it matters?

Cadair: Well that's the bit that matters at the moment, but the whole file must be parseable on py2. i.e. no f-strings or other fun things.

astrofrog: Ok cool - can we add some kind of CI test to make sure it stays that way?

Cadair: maaayybbeee?! :stuck_out_tongue:

Cadair:
@astrofrog little test added.

PR #427, opened 2019-09-01
Fix edit-on-github

mhvk:
Use the new way of using `import_module` instead of `__import__`, and fixes a remaining 'package_name'.

Found this while updating my own packages following the new template. It doesn't seem the `edit_on_github` functionality is tested...


zonca:
hit this same issue today, is it possible to merge this?

PR #433, opened 2019-09-19
Fix instructions to update via git

zonca:


docs/updating.rst
```diff
@@ -62,12 +62,12 @@ this branch.
...
 
 2. Pull from the remote::
 
-   $ git remote update
+   $ git fetch package-template
 
 3. Merge the branch::
 
-   $ git merge package-template/rendered
+   $ git merge package-template/master
```
pllim: I thought `master` is defunct?

zonca: there is no `rendered` branch, and `master` looks like the correct rendered template:
https://github.com/astropy/package-template/branches

PR #438, opened 2020-01-17
Refactor package template to follow APE 17

astrofrog:
:construction_worker_man: This refactors the package template to follow the recommendations in [APE 17](https://github.com/astropy/astropy-APEs/blob/master/APE17.rst). Specifically, the changes are:

* Added the APE 17 migration guide to the docs
* Removed anything to do with astropy-helpers
* Added a ``pyproject.toml`` file
* Updated ``setup.py`` to match what we have in astropy core
* Added a ``tox.ini`` configuration file
* Updated Travis CI configuration to use tox
* Make use of setuptools-scm
* Removed coveragerc in favor of having the coverage config in setup.cfg
* Updated ``conftest.py``

In addition I have made the following changes:

* Added the ``use_compiled_extensions`` option for cookiecutter, because now there are some changes that are required for compiled extensions even if there is no example code.
* Removed ``include_example_cython_code`` and instead rely on ``use_compiled_extensions`` used together with ``include_example_code``
* Removed the ``long_description`` question in favor of using the ``README.rst`` as is often done
* Removed the automodapi templates which should no longer be needed (not sure why we still had them)

:book: To see a preview of the migration guide, see https://external-builds.readthedocs.io/html/astropy-package-template/438/index.html (or click 'Details' in the status checks for the RTD build)

:warning: If you want to try out the guide, in Step 0, use the following to use the template from this PR:

```
cookiecutter gh:astrofrog/package-template --checkout infrastructure-refactor -o my_package_tmp
```

There is still some cleanup I'd like to make to the template unrelated to APE 17, so I'll leave this to another PR. In particular, @Cadair has been doing some work creating a more generic non-astropy-specific template and as part of that has developed a nice way to test the template, but that can wait for another PR.

Preview of the rendered template: https://github.com/astrofrog/package-template/tree/preview-ape17


astrofrog:
(not actually ready for review, but just want to make sure the CI works correctly)

astrofrog:
This is now ready for review!

hooks/post_gen_project.py
```diff
@@ -47,15 +47,12 @@ def process_licence(licence_name):
...
     if '{{ cookiecutter.include_example_code }}' != 'y':
         remove_dir('{{ cookiecutter.module_name }}/example_subpkg/')
         remove_file('{{ cookiecutter.module_name }}/example_mod.py')
         remove_file('{{ cookiecutter.module_name }}/tests/test_example.py')
 
-    if '{{ cookiecutter.include_example_cython_code }}' != 'y':
+    if '{{ cookiecutter.use_compiled_extensions }}' != 'y' or '{{ cookiecutter.include_example_code }}' != 'y':
         remove_file('{{ cookiecutter.module_name }}/example_c.pyx')
 
     if '{{ cookiecutter.initialize_git_repo }}' == 'y':
```
Cadair: Do we want to keep this? I pulled it from the OA one, feels more like a documentation problem now.

{{ cookiecutter.package_name }}/setup.py
```diff
@@ -1,33 +1,87 @@
...
+# version, so we catch the generic 'Exception'.
+try:
+    from setuptools_scm import get_version
+    version = get_version(root='..', relative_to=__file__)
+except Exception:
+    version = '{version}'
+""".lstrip()
 
-setup(version=version, cmdclass=cmdclass, **package_info)
+{% if cookiecutter.use_compiled_extensions == 'y' %}
```
Cadair: I have a less repetitive version of this in the OA guide.

hooks/post_gen_project.py
astrofrog: I want to remove it but in a separate PR to keep things from escalating too much

{{ cookiecutter.package_name }}/setup.py
astrofrog: Fixed

astrofrog:
Ok I've now simplified the migration guide a bit to avoid repeating whole files when the file can just be copied over as-is from the generated template without any modifications.

astrofrog:
The migration guide can be previewed in the RTD PR build: https://external-builds.readthedocs.io/html/astropy-package-template/438/ape17.html

{{ cookiecutter.package_name }}/setup.py
astrofrog: The generated version looks like:

```
setup(
    use_scm_version={'write_to': os.path.join('test_package', 'version.py'),
                     'write_to_template': VERSION_TEMPLATE},

    ext_modules=get_extensions()

)
```

though... Any ideas on avoiding that?

Cadair: I think changing `{%` to `{%-` should remove the extra whitespace or something like that.

hooks/post_gen_project.py
Cadair: does it trigger a bunch of doc changes or something? it's easy to remove from this angle.

docs/ape17.rst
```diff
@@ -0,0 +1,537 @@
...
+
+`tox <https://tox.readthedocs.io/en/latest/>`__ is a tool for automating
+commands, which is well suited to e.g. running tests for your package or
+building the documentation. One of the benefits of using tox is that it
+will (by default) create a source distribution for your package and
+install it into a virtual environment before running tests or building
+docs, which means that it will be a good test of whether e.g. you have
+declared the package data correctly.
+
+As a starting point, copy over the ``tox.ini`` file generated in :ref:`Step 0 <step0>`.
```
mwcraig: As above, it would be nice to have a rendered example to point to.

docs/ape17.rst
```diff
@@ -0,0 +1,537 @@
...
+
+Preivously, astropy-helpers expected the coverage configuration to
+be located in ``my_package/tests/coveragerc``. This is now no longer
+necessary, so you can now define the coverage configuration inside
+the ``setup.cfg`` file, which should help reduce the number of files
+to keep track of. Add the following to the bottom of your ``setup.cfg``::
+
+    [coverage:run]
+    omit =
+        my_package/_{{ cookiecutter._parent_project }}_init*
```
mwcraig: Is the migration guide `cookiecutter` generated or is this a typo?

docs/ape17.rst
```diff
@@ -0,0 +1,537 @@
...
+        my_package/conftest.py
+        my_package/*setup_package*
+        my_package/tests/*
+        my_package/*/tests/*
+        my_package/extern/*
+        my_package/version*
+        */my_package/_{{ cookiecutter._parent_project }}_init*
```
mwcraig: Ditto here.

docs/ape17.rst
```diff
@@ -0,0 +1,537 @@
...
+customizations in ``coveragerc`` you can include them here, but otherwise the
+above should be sufficient.
+
+Step 16 - conftest.py file updates
+----------------------------------
+
+For the header in your test runs to be correct with the latest versions of
+astropy, you will need to make sure that you update your ``conftest.py`` file as
+described in the `pytest-astropy-header instructions
+<https://github.com/astropy/pytest-astropy-header#migrating-from-the-astropy-header-plugin-to-pytest-astropy>`_.
```
mwcraig: It would be nice to close https://github.com/astropy/pytest-astropy-header/issues/14 before the migration guide is distributed widely. I found that error particularly confusing.

docs/ape17.rst
```diff
@@ -0,0 +1,537 @@
...
+   my_package.tests = data/*
+
+In the above example, all ``.fits`` and ``.csv`` in the package will be
+included as well as all files inside ``my_package/tests/data``.
+
+Step 5 - Update your ``setup.py`` file
+--------------------------------------
+
+Copy the ``setup.py`` file you generated in :ref:`Step 0 <step0>` and replace your existing one
+- it should be good to go as-is without any further customizations.
```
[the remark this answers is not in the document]
Cadair: The master branch of the repo is the rendered version with defaults, so we can point to that.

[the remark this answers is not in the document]
astrofrog: @mwcraig - the issue is that then the rendered version does need a customization. Even if you already have a package going the cookiecutter step 0 does make things a lot easier. Same goes for the tox file where the package name has to be customized. Maybe I should clarify this in Step 0?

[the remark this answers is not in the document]
astrofrog: The other issue is that the setup.py file is different whether or not there are compiled extensions. So it really is easier to just run cookiecutter yourself in Step 0 and then copy the file over.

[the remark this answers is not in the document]
astrofrog: I've re-worded Step 0 to make it clearer this is not an optional step - see changes in e1658a33dc75102ebcba1078fd1c622409949841. @mwcraig - does that make it clearer?

Just to be clear, I could link to the rendered versions in master, I'm just worried it will add a whole bunch of text to explain the changes that have to be made to them.

docs/ape17.rst
astrofrog: Fixed, thanks!

docs/ape17.rst
astrofrog: Fixed

docs/ape17.rst
astrofrog: See https://github.com/astropy/pytest-astropy-header/pull/16

docs/ape17.rst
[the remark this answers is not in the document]
mwcraig: > I've re-worded Step 0 to make it clearer this is not an optional step - see changes in e1658a3. @mwcraig - does that make it clearer?

Yep, this looks much clearer.

pllim:
The RTD error is due to cache being incompatible with pip 20.0 upgrade. I don't know how to clear the cache in a PR auto builder. Maybe you either ignore it or force a rerun until cache clears itself? Either push commit or close/re-open to force a RTD rerun.

mwcraig:
I think the problem with pip 20.0.0 is bigger than cache: https://github.com/pypa/pip/issues/7620

docs/ape17.rst
```diff
@@ -0,0 +1,533 @@
...
+   git rm ah_bootstrap.py
+
+You can now commit your changes with::
+
+   git commit -m "Remove astropy-helpers submodule"
+
+Step 2: Update/create ``setup.cfg``
+-----------------------------------
+
+The next step is to update make sure that you have a ``setup.cfg`` file
```
pllim: Maybe add a link to official documentation at https://docs.python.org/3/distutils/configfile.html ?

docs/ape17.rst
```diff
@@ -0,0 +1,533 @@
...
+and specify your required dependencies one per line in the
+``install_requires`` section. Make sure the ``python_requires`` line is
+set to indicate the correct minimum Python version for your package.
+
+If you already had a file, make sure you remove the following entries
+(if present):
+
+-  ``package_name``
+-  ``version``
+-  ``setup_requires``
```
pllim: Removing `setup_requires` directly contradicts what is in Step 8.

docs/ape17.rst
```diff
@@ -0,0 +1,533 @@
...
+
+If you have Cython extensions or your extensions use the NumPy C API,
+proceed to Step 7c, otherwise you can proceed to Step 8.
+
+Step 7b - Using extension-helpers
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can use the extension-helpers package to:
+
+-  Automatically define extensions for Cython files
```
pllim:
```suggestion
-  Automatically define extensions for Cython files.
```

docs/ape17.rst
```diff
@@ -0,0 +1,533 @@
...
+Step 7a - Defining extensions manually
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+You can define extensions manually as described
+`here <https://oa-packaging-guide-preview.readthedocs.io/en/latest/extensions.html#defining-extensions-in-setup-py>`__.
+If you do this, you can remove all ``setup_package.py`` files in your
+package, and you don't need to include extension-helpers in the
+``pyproject.toml`` file.
+
+If you have Cython extensions or your extensions use the NumPy C API,
```
pllim: I think this sentence is duplicated in 7b and it makes more sense living in 7b.
```suggestion
```

docs/ape17.rst
```diff
@@ -0,0 +1,533 @@
...
+proceed to Step 7c, otherwise you can proceed to Step 8.
```
pllim:
```suggestion
```

docs/ape17.rst
```diff
@@ -0,0 +1,533 @@
...
+-  Automatically define extensions for Cython files
+-  Pick up extensions declared in ``setup_package.py`` files, as
+   described in the `extension-helpers
+   documentation <https://extension-helpers.readthedocs.io/en/latest/>`__.
+
+The latter works by looking through all the ``setup_package.py`` files
+in your package and executing the ``get_extensions()`` functions, which
+each should return a list of extensions. Check through your existing
+``setup_package.py`` files (if any), and make sure that any
+``astropy_helpers`` imports are changed to ``extension_helpers``.
```
pllim: If I remember correctly, a simple search and replace is not enough. You also moved some API up.

There is also a change from string to function call that has something to do with Numpy `get_include`.

docs/ape17.rst
```diff
@@ -0,0 +1,533 @@
...
+Whenever a new major Python version is released, you will likely need to
+update the Cython pinning to use the most recent Cython version available.
+
+Step 8 - Using setuptools_scm
+-----------------------------
+
+The `setuptools_scm <https://pypi.org/project/setuptools-scm/>`__
+package is now recommended to manage the version numbers for your
+package. The way this works is that instead of setting the version
+number manually in e.g. ``setup.cfg`` or elsewhere in your package,
```
pllim:
```suggestion
number manually in, e.g., ``setup.cfg`` or elsewhere in your package,
```

docs/ape17.rst
```diff
@@ -0,0 +1,533 @@
...
+the version number is based on git tags.
+
+In Steps 5 and 6, we already added the required entry for
```
pllim: Use `ref` links for the Steps 5 and 6 mentioned?

docs/ape17.rst
```diff
@@ -0,0 +1,533 @@
...
+   prune astropy_helpers/astropy_helpers/tests
+
+Then add a new line near the top with the following::
+
+   include pyproject.toml
+
+Step 11 - Updating your documentation configuration
+---------------------------------------------------
+
+You will need to edit the ``docs/conf.py`` file to make sure it does not
```
pllim: Also mention the change needed to grab package version number?

docs/ape17.rst
```diff
@@ -0,0 +1,533 @@
...
+   conf = ConfigParser()
+
+Find and replace any instances of ``package_name`` in the file with
+``name``.
+
+Step 12 - Setting up tox
+------------------------
+
+`tox <https://tox.readthedocs.io/en/latest/>`__ is a tool for automating
+commands, which is well suited to e.g. running tests for your package or
```
pllim:
```suggestion
commands, which is well suited to, e.g., running tests for your package or
```

docs/ape17.rst
```diff
@@ -0,0 +1,533 @@
...
+building the documentation. One of the benefits of using tox is that it
+will (by default) create a source distribution for your package and
+install it into a virtual environment before running tests or building
+docs, which means that it will be a good test of whether e.g. you have
```
pllim:
```suggestion
docs, which means that it will be a good test of whether, e.g., you have
```

docs/ape17.rst
```diff
@@ -0,0 +1,533 @@
...
+If you use Travis CI, a good place to start is the ``.travis.yml`` file
+generated in :ref:`Step 0 <step0>`, and you can then see if any previous customizations you had
+made need to be copied over. This file shows how one can configure Travis to use
+tox, optionally using conda via ci-helpers to set up Python on MacOS X and
+Windows.
+
+Step 14 - Update ReadTheDocs configuration
+------------------------------------------
+
+With the set-up described in this migration guide, you should be able to
```
pllim: Add a link to https://docs.readthedocs.io/en/stable/config-file/v2.html

docs/ape17.rst
```diff
@@ -0,0 +1,533 @@
...
+simplify the configuration for ReadTheDocs. This can be done via a
+``readthedocs.yml`` or ``.readthedocs.yml`` file (the latter is recommended).
+You should be able to copy over the ``.readthedocs.yml`` file created in :ref:`Step 0 <step0>`.
+With this updated file, you should now be able to remove any pip requirements
+file or conda yml file that were previously used by ``readthedocs.yml``.
```
pllim:
```suggestion
file or conda YAML file that were previously used by ``readthedocs.yml``.
```

docs/ape17.rst
```diff
@@ -0,0 +1,533 @@
...
+above should be sufficient.
+
+Step 16 - conftest.py file updates
+----------------------------------
+
+For the header in your test runs to be correct with the latest versions of
+astropy, you will need to make sure that you update your ``conftest.py`` file as
+described in the `pytest-astropy-header instructions
+<https://github.com/astropy/pytest-astropy-header#migrating-from-the-astropy-header-plugin-to-pytest-astropy>`_.
+You can also just copy over the file created in :ref:`Step 0 <step0>` and add back any
```
pllim:
```suggestion
You can also copy over the file created in :ref:`Step 0 <step0>` and add back any
```

docs/ape17.rst
```diff
@@ -0,0 +1,533 @@
...
+
+-  ``[build_docs]``
+-  ``[build_sphinx]``
+-  ``[upload_docs]``
+
+You should also add ``pip-wheel-metadata`` to your ``.gitignore`` file.
+
+**Once you are done, if you would like us to help by reviewing your changes,
+you can open a pull request to your package and mention @astrofrog or
+@Cadair to ask for a review**
```
pllim:
```suggestion
@Cadair to ask for a review.**
```

{{ cookiecutter.package_name }}/setup.cfg
```diff
@@ -1,62 +1,74 @@
...
 console_scripts =
-    {% if cookiecutter.include_example_code == 'y' %}
     astropy-package-template-example = packagename.example_mod:main
-    {% else %}
-    # astropy-package-template-example = packagename.example_mod:main
-    {% endif %}
+{% endif %}
+
+[options.extras_require]
+all =
```
pllim: Why the empty `all=`?

{{ cookiecutter.package_name }}/setup.cfg
```diff
@@ -1,62 +1,74 @@
...
+    except ImportError
+    # Don't complain if tests don't hit assertions
+    raise AssertionError
+    raise NotImplementedError
+    # Don't complain about script hooks
+    def main\(.*\):
+    # Ignore branches that don't pertain to this version of Python
+    pragma: py{ignore_python_version}
+    # Don't complain about IPython completion helper
+    def _ipython_key_completions_
```
pllim: Why is IPython helper a problem for coverage test? Do we really need this?

{{ cookiecutter.package_name }}/{{ cookiecutter.module_name }}/conftest.py
```diff
@@ -3,47 +3,38 @@
 import os
 
 from astropy.version import version as astropy_version
+
+# For Astropy 3.0 and later, we can use the standalone pytest plugin
 if astropy_version < '3.0':
```
pllim: Do we really need the `if` here? Does `pytest_astropy_header` not work with `astropy<3`?

{{ cookiecutter.package_name }}/{{ cookiecutter.module_name }}/conftest.py
```diff
@@ -3,47 +3,38 @@
...
+    # Customize the following lines to add/remove entries from the list of
+    # packages for which version numbers are displayed when running the tests.
+    PYTEST_HEADER_MODULES.pop('Pandas', None)
+    PYTEST_HEADER_MODULES['scikit-image'] = 'skimage'
+
+    from .version import version
+    packagename = os.path.basename(os.path.dirname(__file__))
+    TESTED_VERSIONS[packagename] = version
+
+from astropy.tests.helper import enable_deprecations_as_exceptions  # noqa
```
pllim: This should also be commented. Otherwise, there is an unused import.

.readthedocs.yml
```diff
@@ -1,4 +1,11 @@
...
+version: 2
+
+build:
+  image: latest
 
 python:
-   version: 3.5
+  version: 3.7
+  install:
+    - requirements: requirements.txt
```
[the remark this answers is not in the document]
astrofrog: These are the requirements to build the docs for the package template, not the generated package - there is no setup.cfg at the root of package-template.

QUESTIONS.rst
```diff
@@ -5,20 +5,19 @@ To use the package template run ``cookiecutter gh:astropy/package-template``,
...
-13. ``project_version``: The current version number of your project.
-14. ``use_travis_ci``: If 'y' the template will include an example ``.travis.yml`` file for the Travis CI service.
-15. ``use_read_the_docs``: If 'y' the ``read_the_docs.yml`` and ``.rtd-environment.yml`` files will be included for using conda on Read the Docs.
-16. ``sphinx_theme``: The value of the ``html_theme`` variable in the sphinx configuration file.
-17. ``minimum_python_version``: Version string of minimum supported Python version
+13. ``use_travis_ci``: If ``'y'`` the template will include an example ``.travis.yml`` file for the Travis CI service.
+14. ``use_read_the_docs``: If ``'y'`` the ``read_the_docs.yml`` and ``.rtd-environment.yml`` files will be included for using conda on Read the Docs.
+15. ``sphinx_theme``: The value of the ``html_theme`` variable in the sphinx configuration file.
+16. ``initialize_git_repo``: If `gitpython <https://gitpython.readthedocs.io/en/stable/>`_ is installed this option will turn the rendered package into a git repository and add and initilize the ``astropy_helpers`` submodule.
```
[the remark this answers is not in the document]
astrofrog: I forgot to remove that :) I renumbered the questions hence why it looks like I added these lines. In fact I'm going to go with @Cadair's suggestion of removing the git option altogether since it's not that useful without astropy-helpers.

QUESTIONS.rst
```diff
@@ -5,20 +5,19 @@ To use the package template run ``cookiecutter gh:astropy/package-template``,
 once you have run this command you will be asked a series of questions. Below is
 a description of each of the prompts.
 
-1. ``package_name``: This is a human readable name for your package, like 'Astropy' or 'SunPy'.
-2. ``module_name``: This is the name of your python package i.e. 'astropy' or 'sunpy'.
+1. ``package_name``: This is a human readable name for your package, like ``Astropy`` or ``SunPy``.
+2. ``module_name``: This is the name of your python package i.e. ``astropy`` or ``sunpy``.
 3. ``short_description``: This is a one sentence description of your package.
-4. ``long_description``: This is a multi-line description of your package.
```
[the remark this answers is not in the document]
astrofrog: I think twine chokes if the README is not valid rst, but otherwise I've never seen issues with complexity. I do think we want to encourage this approach since it produces much nicer pages on PyPI.

cookiecutter.json
```diff
@@ -19,9 +17,5 @@
     "initialize_git_repo": "y",
     "_parent_project": "astropy",
     "_install_requires": "astropy",
-    "_copy_without_render": [
-        "docs/_templates"
-    ],
-    "minimum_python_version": ["3.7", "3.6", "3.5"],
-    "astropy_helpers_version": "v3.2.1"
+    "minimum_python_version": ["3.7", "3.6", "3.5"]
```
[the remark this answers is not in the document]
astrofrog: Done

docs/ape17.rst
```diff
@@ -0,0 +1,533 @@
+APE 17 Migration Guide
+======================
+
+.. warning:: This guide is not yet ready for widespread use and may
```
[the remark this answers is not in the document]
astrofrog: Yes - we still shouldn't advertise it widely until all coordinated packages have been migrated. Once this is merged, we should go through and systematically update all coordinated packages (even if already updated back in December, to make sure they are all consistent). And once that's done and everything is working we can remove this warning and advertise. Does that seem ok?

docs/ape17.rst
```diff
@@ -0,0 +1,533 @@
...
+             still change significantly.
+
+The Astropy project is now transitioning from using astropy-helpers for
```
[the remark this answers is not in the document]
astrofrog: This sentence refers to the project, but I just added a sentence at the end of the paragraph saying *The core astropy package has already migrated and these changes will be included
in astropy v4.1.*

docs/ape17.rst
astrofrog: Done

docs/ape17.rst
astrofrog: I've remove it from the list, but I've added back a mention to remove any existing ``setup_requires`` entry and that we'll add it back in Step 8. The point here is that any existing setup_requires with e.g. numpy or cython is no longer necessary.

docs/ape17.rst
astrofrog: I'd like to keep this, because people will read *either* 7a or 7b (but won't do both) so it's important to mention this in both cases.

docs/ape17.rst
astrofrog: Fixed!

docs/ape17.rst
astrofrog: Done! I also just did refs for all step links

docs/ape17.rst
astrofrog: Done

{{ cookiecutter.package_name }}/setup.cfg
astrofrog: It's easier this way to show where optional dependencies should go.

{{ cookiecutter.package_name }}/setup.cfg
astrofrog: Packages can include this method on classes to help with IPython tab completion which isn't something that can be tested with pytest easily - I was just copying from the core package.

{{ cookiecutter.package_name }}/{{ cookiecutter.module_name }}/conftest.py
astrofrog: This is following the setup recommended in https://github.com/astropy/pytest-astropy-header#migrating-from-the-astropy-header-plugin-to-pytest-astropy - basically the plugin indeed doesn't fully work with astropy < 3 so in that case we need to use the plugin that was bundled with astropy core (basically for astropy < 3 we can't disable the built-in one so we can't use the standalone plugin too)

{{ cookiecutter.package_name }}/{{ cookiecutter.module_name }}/conftest.py
astrofrog: Done

docs/ape17.rst
astrofrog: This doesn't actually need to change compared to before for everything to work - but I want to do a follow up PR cleaning up the ``conf.py`` file so we can maybe discuss what our preferred approach is there?

hooks/post_gen_project.py
astrofrog: I decided to remove it in the end

QUESTIONS.rst
```diff
@@ -5,20 +5,18 @@ To use the package template run ``cookiecutter gh:astropy/package-template``,
...
+9. ``use_compiled_extensions``: Whether you plan to use compiled extensions in your package
 10. ``include_cextern_folder``: The cextern folder should be used if you are including non-python C code.
-11. ``edit_on_github_extension``: Set to "True" to enable the edit on GitHub sphinx extension.
+11. ``edit_on_github_extension``: Set to ``True`` to enable the edit on GitHub sphinx extension.
 12. ``github_project``: This is the GitHub identifier for the edit on GitHub extension and the changelog link extension.
-13. ``project_version``: The current version number of your project.
-14. ``use_travis_ci``: If 'y' the template will include an example ``.travis.yml`` file for the Travis CI service.
-15. ``use_read_the_docs``: If 'y' the ``read_the_docs.yml`` and ``.rtd-environment.yml`` files will be included for using conda on Read the Docs.
-16. ``sphinx_theme``: The value of the ``html_theme`` variable in the sphinx configuration file.
-17. ``minimum_python_version``: Version string of minimum supported Python version
```
bsipocz: why ditch the minimum python version?

{{ cookiecutter.package_name }}/tox.ini
```diff
@@ -0,0 +1,82 @@
...
+    numpy117: with numpy 1.17.*
+    numpy118: with numpy 1.18.*
+
+deps =
+
+    numpy116: numpy==1.16.*
+    numpy117: numpy==1.17.*
+    numpy118: numpy==1.18.*
+
+    astropy30: astropy==3.0.*
```
bsipocz: not exactly clear where you use these, don't you need to list it above in `envlist` as a pattern?

{{ cookiecutter.package_name }}/setup.cfg
bsipocz: can you leave a comment for it then?

bsipocz: also, one could have the deps listed in the cookiecutter template?

mhvk:
@astrofrog - is the rendered version up-to-date? I ask in part since it mentions `gitpython` which you seem to be removing above (though perhaps I misread that comment!).

lglattly:
> I am not sure if @lglattly's work includes copyediting outside the core library. If yes, then maybe she could have a look at `ape17.rst` too?

Yes I can work on copy editing outside of the core library! @adrn can help get me set up with other repositories as needed. I can take a look at `ape17.rst`, but maybe it makes sense for me to do an editing pass once it's merged and open a copy editing PR for it? The addition looks long and it might be too annoying to deal with lots of small copy editing suggestions here. Let me know!

bsipocz:
> maybe it makes sense for me to do an editing pass once it's merged and open a copy editing PR for it

I think that's a great idea. As I see the plan is to get this merged, but don't over advertize it until the coordinated libraries are transitioned, and frankly that will take time anyway.

{{ cookiecutter.package_name }}/setup.cfg
astrofrog: Do you mean this should be a question cookiecutter asks?

QUESTIONS.rst
astrofrog: This was a mistake - it was still in the ``cookiecutter.json`` file but I had accidentally removed it from the list of options/questions. Thanks for catching this! Fixed.

{{ cookiecutter.package_name }}/tox.ini
astrofrog: You don't need to list all possible options in ``envlist`` - you can basically include ``astropy30`` in the env name anywhere if you want to pin it. So you could do e.g. ``py38-test-astropy30`` or ``py38-numpy117-astropy30-test`` (the order is not important). I can add them to the envlist though for clarity.

{{ cookiecutter.package_name }}/setup.cfg
bsipocz: yes, but it's not a strong preference, just along the line of the logic that we otherwise use the cookiecutter to fill out these boilerplates all the other places.

{{ cookiecutter.package_name }}/tox.ini
bsipocz: ahh, ok. This just shows that I should have looked into tox a bit more.

{{ cookiecutter.package_name }}/setup.cfg
astrofrog: I've now added this - both for required and optional dependencies

astrofrog:
@mhvk - no the rendered version on master only gets updated when PRs like this are merged. But I've pushed a rendered version to a branch in my fork so you can see it if you like: https://github.com/astrofrog/package-template/tree/preview-ape17

{{ cookiecutter.package_name }}/setup.py
astrofrog: Ok that worked - thanks!

docs/ape17.rst
astrofrog: See previous comment - I've now clarified Step 0 to make it clear it's useful even for existing packages.

docs/ape17.rst
astrofrog: https://github.com/astropy/pytest-astropy-header/pull/16 has been merged, so resolving this

docs/ape17.rst
astrofrog: See previous comment

astrofrog:
I think I've addressed most comments - I've resolved ones that are clear cut and left others marked as unresolved.

@pllim, you said:

> No mention of tagging dev for branch releases?

I don't think that belongs here - release instructions for packages (including packages using the template) are in the astropy core docs for now. I'd like to consolidate that more but for now this isn't the right place I think.

{{ cookiecutter.package_name }}/.gitignore
```diff
@@ -55,6 +55,7 @@ distribute-*.tar.gz
 .project
 .pydevproject
 .settings
+pip-wheel-metadata
```
pllim: I am not sure if a trailing slash is compulsory when it is a directory but I do it anyway for my packages.
```suggestion
pip-wheel-metadata/
```

astrofrog:
@pllim - thanks! I've implemented both comments.

mhvk:
@astrofrog - am doing it for my baseband package now and have some comments - mostly that the cookiecutter results can be highlighted more - it does almost everything!

mhvk:
p.s. Full review later today.

bsipocz:
@astrofrog - I won't have time to do the migration with a package right now, so I  don't expect to have any more comments than the ones you already addressed

astrofrog:
@bsipocz - ok sounds good!

@mhvk - great, looking forward to it! :)

pllim:
@mhvk , I prefer to have `setup.cfg` explanations even when `cookiecutter` did the work, as not everyone would like to re-run the template and move things around, especially if the existing package has other customization involved.

{{ cookiecutter.package_name }}/tox.ini
```diff
@@ -0,0 +1,87 @@
...
+    pip >= 19.3.1
+isolated_build = true
+
+[testenv]
+
+# Pass through the following environemnt variables which may be needed for the CI
+passenv = HOME WINDIR LC_ALL LC_CTYPE CC CI TRAVIS
+
+# Run the tests in a temporary directory to make sure that we don't import
+# astropy from the source tree
```
mhvk: astropy -> package name

docs/ape17.rst
```diff
@@ -0,0 +1,562 @@
...
+
+As a starting point, copy over the ``tox.ini`` file generated in :ref:`Step 0 <step-rerender>`.
+You can always then customize it if needed (although it should work out
+of the box).
+
+Once you have done this you should be able to do the following:
+
+Run tests with minimal dependencies::
+
+   tox -e test
```
mhvk: Trying this it seems to automatically run coverage as well; not sure if that's intended.

tepickering:
i'm still in the process of my migrating my repos, but the changes here have been a HUGE help! thanks! the one issue i had was with package data being defined in both `setup.cfg` and `MANIFEST.in`. my understanding of best practices is that `include_package_data = True` should be set in `[options]` in `setup.cfg` and then `MANIFEST.in` used to configured everything that should be included. indeed, this is what i need to do to reliably package my nested directories of package and test data. is there a counter-argument for using `[options.package_data]` instead?

{{ cookiecutter.package_name }}/tox.ini
```diff
@@ -0,0 +1,87 @@
...
+    devdeps: git+https://github.com/astropy/astropy.git#egg=astropy
+
+# The following indicates which extras_require from setup.cfg will be installed
+extras =
+    test
+    alldeps: all
+
+commands =
+    pip freeze
+    pytest --pyargs {{ cookiecutter.module_name }} {toxinidir}/docs --cov {{ cookiecutter.module_name }} --cov-config={toxinidir}/setup.cfg {posargs}
```
mhvk: Of course, coverage is listed here, so it always happens.

mhvk:
@pllim - very much agreed! So, really all what I suggested is to mention that one can glean a lot of the needed changes from the files generated in step 0.

{{ cookiecutter.package_name }}/tox.ini
mhvk: I guess it needs the following from astropy's tox.ini:
```
    !cov: pytest --pyargs packagename {toxinidir}/docs {posargs}
```

{{ cookiecutter.package_name }}/{{ cookiecutter.module_name }}/conftest.py
```diff
@@ -3,47 +3,37 @@
...
-## 'setuptools'), add:
-##     modules_to_ignore_on_import=['module_1', 'module_2']
-## To ignore some specific deprecation warning messages for Python version
-## MAJOR.MINOR or later, add:
-##     warnings_to_ignore_by_pyver={(MAJOR, MINOR): ['Message to ignore']}
-# enable_deprecations_as_exceptions()
+    from pytest_astropy_header.display import PYTEST_HEADER_MODULES, TESTED_VERSIONS
+
+
+def pytest_configure(config):
```
mhvk: Not sure this is working properly - I'm seeing `Pandas` listed as not available, instead of simply omitted.

mhvk: Hmm, since `astropy_helpers` is still listed, there might be some out-of-date cruft getting in the way.

mhvk:
With `tox -e build_docs`, I get the complaint:
```
Configuration error:
There is a programmable error in your configuration file:

Traceback (most recent call last):
  File "/data/mhvk/packages/baseband/.tox/build_docs/lib/python3.7/site-packages/sphinx/config.py", line 368, in eval_config_file
    execfile_(filename, namespace)
  File "/data/mhvk/packages/baseband/.tox/build_docs/lib/python3.7/site-packages/sphinx/util/pycompat.py", line 81, in execfile_
    exec(code, _globals)
  File "/data/mhvk/packages/baseband/docs/conf.py", line 162, in <module>
    if versionmod.release:
AttributeError: module 'baseband.version' has no attribute 'release'
```
Also in the cookiecutter template, there seems an expectation of a `version.release` attribute, inside the `edit_on_github` part.

pllim:
> AttributeError: module 'baseband.version' has no attribute 'release'

The fix:

```python
from pkg_resources import get_distribution

# The full version, including alpha/beta/rc tags.
release = get_distribution(project).version
# The short X.Y version.
version = '.'.join(release.split('.')[:2])
```

In my review, I asked for this to be added to the guide, but @astrofrog wanted to save it for a separate clean-up PR.

mhvk:
@pllim - should that go in `conf.py`? Or rather in the `VERSION_TEMPLATE` inside `setup.py`?

pllim:
@mhvk , that should be inside `docs/conf.py`. Sorry for the confusion!

mhvk:
Sorry, getting more confused: there are already definitions of `release` and `version` inside `docs/conf.py` - those should be ignored here?

https://github.com/astrofrog/package-template/blob/ef9573da29fc3b7ee7e296d308da09dabaf7993d/docs/conf.py#L50-L53

mhvk:
I'm actually confused - that `conf.py` I linked to looks very substantially different from the one `cookiecutter` made. I better check more carefully....

mhvk:
OK, I just looked at the wrong file: https://github.com/astrofrog/package-template/blob/ef9573da29fc3b7ee7e296d308da09dabaf7993d/docs/conf.py#L50-L53

Funny enough the last commit is for "fix edit on github" :smile:

pllim:
@mhvk , now I am confused. Did you end up having to patch with https://github.com/astropy/package-template/pull/438#issuecomment-579969637 or everything works from the template?

mhvk:
@pllim - I was confused too, as I was looking at the wrong file. I was wondering about your patch since that defines `release` and `version`, but those are [already defined in `conf.py`](https://github.com/astrofrog/package-template/blob/ef9573da29fc3b7ee7e296d308da09dabaf7993d/%7B%7B%20cookiecutter.package_name%20%7D%7D/docs/conf.py#L82-L85); what is missing is a `version.release`. So, to apply your patch, do I just let that replace the `version` and `release` that are in `conf.py` and then use those instead of `versionmod.release`? (latter at https://github.com/astrofrog/package-template/blob/ef9573da29fc3b7ee7e296d308da09dabaf7993d/%7B%7B%20cookiecutter.package_name%20%7D%7D/docs/conf.py#L159-L170)

pllim:
@mhvk ... Oh, I get it now. For that section in particular, maybe you should only provide the "master" option if you follow the same release procedures as astropy core library. See astropy/astropy#9379

I don't use "edit on GitHub" for my stuff, so my patch is probably irrelevant to that logic block. Sorry for the confusion! :see_no_evil:

mhvk:
@astrofrog - separately, in recommending `tox`, it is perhaps good for people to realize that this means you get a `.tox` subdirectory for every package you have that will contain GBs of data. It would be nice to isolate this from the package directory (I'll see if I can find the relevant environment variables or so). Similarly, any temporary files are now stored in `.tmp` rather than on a `/tmp` partition, ie., more stuff to clean up oneself (of course, only a `git clean` away).

bsipocz:
@mhvk - thanks for digging up these issues.

mhvk:
@pllim - yes, that makes sense. Much better to always point to master. since if things have already changed, you really want people to notice before doing work!

docs/ape17.rst
astrofrog: I had deliberately just enabled it by default - for astropy core the performance penalty is large, but for most other packages I thought it might make sense to just always do it. But happy to also go with a set-up similar to in the core package if people would prefer (where coverage is opt-in with ``-cov``).

mhvk: Indeed, for my project, it is not a big deal, but I think then you also need to adjust the `tox.ini` to remove the `-cov` option. So it may be simpler to just make it the same. Which for me is also better - ideally everything is essentially the same as for astropy.

{{ cookiecutter.package_name }}/{{ cookiecutter.module_name }}/conftest.py
mhvk: Weird. I tried just adding a print statement inside `conftest.pytest_configure` and it only gets shown inside `collecting...` - indeed, if I just add a breakpoint inside that function, the break also happens after the header is printed. 

If I instead add a new `conftest.py` in my root directory, the removals from `PYTEST_HEADER_MODULES` do work (though for some reason it always list astropy_helpers).

So, here, it may be most relevant to have a `conftest.py` in the root directory; the rest may be more of an issue for `pytest_astropy_header`

mhvk: And for `pytest_astropy_header`, the present of `astropy-helpers` is of course already addressed in https://github.com/astropy/pytest-astropy-header/pull/16

mhvk: OK, the following `conftest.py` in the root directory works at least for me:
```
import os

try:
    from pytest_astropy_header.display import PYTEST_HEADER_MODULES, TESTED_VERSIONS
except ImportError:
    PYTEST_HEADER_MODULES = {}
    TESTED_VERSIONS = {}


def pytest_configure(config):

    config.option.astropy_header = True

    # Customize the following lines to add/remove entries from the list of
    # packages for which version numbers are displayed when running the tests.
    PYTEST_HEADER_MODULES.pop('Pandas', None)
    PYTEST_HEADER_MODULES.pop('h5py', None)
    PYTEST_HEADER_MODULES.pop('Matplotlib', None)
    PYTEST_HEADER_MODULES.pop('Scipy', None)
    PYTEST_HEADER_MODULES.pop('astropy-helpers', None)
    PYTEST_HEADER_MODULES['Astropy'] = 'astropy'

    from baseband.version import version
    packagename = os.path.basename(os.path.dirname(__file__))

    TESTED_VERSIONS[packagename] = version

# Uncomment the last two lines in this block to treat all DeprecationWarnings as
# exceptions. For Astropy v2.0 or later, there are 2 additional keywords,
# as follow (although default should work for most cases).
# To ignore some packages that produce deprecation warnings on import
# (in addition to 'compiler', 'scipy', 'pygments', 'ipykernel', and
# 'setuptools'), add:
#     modules_to_ignore_on_import=['module_1', 'module_2']
# To ignore some specific deprecation warning messages for Python version
# MAJOR.MINOR or later, add:
#     warnings_to_ignore_by_pyver={(MAJOR, MINOR): ['Message to ignore']}
from astropy.tests.helper import enable_deprecations_as_exceptions  # noqa
enable_deprecations_as_exceptions()
```

bsipocz: would the `    config.option.astropy_header = True` work if you ended up in the except part of the top try/except? In a similar situation I simply put the whole stuff in the try part.

mhvk:
Another issue with tests is that any files created in the doctests are just put wherever pytest is run. But I now see that that is just https://github.com/astropy/pytest-doctestplus/issues/68 - would be nice to address that!

mhvk:
Just to summarize, since I left rather many comments: the real ones are in-line:
- move `conftest.py` up one level to ensure that the header actually displays what is wanted (need to be sure to get `version` correctly).
- Make handling of coverage identical to that in astropy.
- Some small cleanups in ape17.rst and tox.ini

{{ cookiecutter.package_name }}/{{ cookiecutter.module_name }}/conftest.py
bsipocz: If you already have `baseband` in this line: `from baseband.version import version`, then why do the next two as a hackary for packagename, etc?

mhvk: Ah, yes, that was rather silly! It does seem that ideally one would just have this stuff in `setup.cfg`.

docs/ape17.rst
```diff
@@ -0,0 +1,562 @@
...
+------------------------------
+
+To remove the astropy-helpers submodule, first, run::
+
+   git rm -r astropy_helpers
+
+if ``astropy_helpers`` was the only submodule in your repository, the
+``.gitmodules`` file will be empty, you can remove this if you wish::
+
+    git rm .gitmodules
```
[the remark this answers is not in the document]
astrofrog: Done

docs/ape17.rst
```diff
@@ -0,0 +1,562 @@
...
+   git rm ah_bootstrap.py
+
+You can now commit your changes with::
+
+   git commit -m "Remove astropy-helpers submodule"
+
+Step 2: Update/create ``setup.cfg``
+-----------------------------------
+
+The next step is to update make sure that you have a `setup.cfg
```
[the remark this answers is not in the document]
astrofrog: I've added a mention of the generated file

docs/ape17.rst
```diff
@@ -0,0 +1,562 @@
...
+
+if ``astropy_helpers`` was the only submodule in your repository, the
+``.gitmodules`` file will be empty, you can remove this if you wish::
+
+    git rm .gitmodules
+
+Next you should remove the ``ah_bootstrap.py`` file::
+
+   git rm ah_bootstrap.py
+
```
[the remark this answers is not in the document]
astrofrog: I added this but in Step 15 not here since this is mainly about astropy-helpers

docs/ape17.rst
```diff
@@ -0,0 +1,562 @@
...
+Step 6: add a ``pyproject.toml`` file
+-------------------------------------
+
+The ``pyproject.toml`` file is used to declare dependencies needed to run
+``setup.py`` and build the package. Copy the ``pyproject.toml`` file you
+generated in :ref:`Step 0 <step-rerender>` and replace your existing one.
+
+If your package doesn’t have any compiled extensions, the file should contain:
+
+.. code:: toml
```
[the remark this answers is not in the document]
astrofrog: Fixed

docs/ape17.rst
```diff
@@ -0,0 +1,562 @@
...
+In :ref:`Steps 5 <step-setup-py>` and :ref:`6 <step-pyproject-toml>`, we already
+added the required entry for setuptools_scm to ``setup.py`` and
+``pyproject.toml``.
+
+In addition to these, we recommend that you define ``setup_requires`` inside the
+``[options]`` section of your ``setup.cfg`` file::
+
+   [options]
+   ...
+   setup_requires = setuptools_scm
```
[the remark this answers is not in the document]
astrofrog: Added a mention of this

docs/ape17.rst
```diff
@@ -0,0 +1,562 @@
...
+file. The git hash is now contained in the version number, so this is no
+longer needed.
+
+Step 9 - Configuring pytest
+---------------------------
+
+To make sure that pytest works properly, you can set a few options in a
+``[tool:pytest]`` section in your ``setup.cfg`` file::
+
+   [tool:pytest]
```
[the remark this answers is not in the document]
astrofrog: Added a mention of this

{{ cookiecutter.package_name }}/tox.ini
astrofrog: changed to 'this package' to just keep it generic

docs/ape17.rst
astrofrog: I've now made it consistent with astropy core and made the ``cov`` option work

{{ cookiecutter.package_name }}/tox.ini
astrofrog: Done

{{ cookiecutter.package_name }}/{{ cookiecutter.module_name }}/conftest.py
astrofrog: I'll comment in the main thread since this is a big-ish issue

astrofrog:
@mhvk - thanks for the careful review!

Regarding the '**release and version in conf.py**, the latest version in the present ``conf.py`` in the template should work nicely:

```
# The short X.Y version.
version = package.__version__.split('-', 1)[0]
# The full version, including alpha/beta/rc tags.
release = package.__version__
```

or alternatively what @pllim said, but indeed I'd like to wait until a follow up PR for that since there's nothing actually wrong with the above approach and in fact it can be faster than using ``get_distribution``. But either way the important thing is that ``release`` is something that is only defined in ``conf.py``, not something the ``package/version.py`` file would ever provide.

Regarding **conftest.py**, things are a little trickier. The issue with moving ``conftest.py`` to the root is that it does not get installed, so then running ``packagename.test()`` in an interactive Python session won't work. In fact, the cleanest approach here is to leave ``conftest.py`` inside the package but then to also duplicate the list of packages to show in the ``setup.cfg`` file:

```
[tool:pytest]
...
astropy_header = true
astropy_header_packages =
    scipy
    matplotlib
```

In general I agree that this is not ideal and really moving ``conftest.py`` to the root of the package would be the 'proper' way to do it and then not supporting running self-tests inside installed packages (i.e. the ``test()`` command). But I'm pretty sure a number of people feel strongly about keeping that.

One alternative I want to put out there is to simply get rid of the list of dependencies in the pytest header and instead just rely on commands like ``pip freeze`` or ``conda list`` if people need to check installed packages. In fact that's why I added ``pip freeze`` as a command inside the tox config, because that gives a much more complete picture of things. This is my personal preference - to keep the custom header for the other bits but to not actually show the package list.

My second favorite option is to actually just generate that list of dependencies automatically from the declared package dependencies, as I suggested in https://github.com/astropy/astropy/issues/8867 - that idea met a lot of opposition but maybe it would be a good time to reconsider. This would simplify things a lot since there would be no need to override the options in either ``conftest.py`` or ``setup.py``.

mhvk:
@astrofrog - I agree that the lines in `conf.py` are fine, but those are not used in `edit_on_github`:

https://github.com/astrofrog/package-template/blob/970fe945438539c1cfb0306088639e5840c75b25/%7B%7B%20cookiecutter.package_name%20%7D%7D/docs/conf.py#L159-L170

That said, I think that @pllim's solution of just setting the relevant branch to `master` is better anyway!

mhvk:
On the list of dependencies, having looked back at the thread, I think the consensus was that there should be a way to specify it manually, but I didn't see a big push against having an automatic default that made sense. To me, the best default would be explicit dependencies mentioned in `setup.cfg`, but that really doesn't matter that much; it could also be something that is a special key, say `astropy_header_packages = 'requierd'` and maybe 'recursive'. 

A propos `package/conftest.py`, could that be generated upon installation like `version.py`?

p.s. It may be smartest to move this to a post-this-PR discussion!

{{ cookiecutter.package_name }}/{{ cookiecutter.module_name }}/conftest.py
```diff
@@ -3,47 +3,37 @@
...
-TESTED_VERSIONS[packagename] = version
-TESTED_VERSIONS['astropy_helpers'] = astropy_helpers_version
+    config.option.astropy_header = True
+
+    # Customize the following lines to add/remove entries from the list of
+    # packages for which version numbers are displayed when running the tests.
+    PYTEST_HEADER_MODULES.pop('Pandas', None)
+    PYTEST_HEADER_MODULES['scikit-image'] = 'skimage'
+
+    from .version import version
```
mhvk: Previously, `version` was guaranteed to exist, but this is no longer true with `setuptools_scm`. To allow people to still run `pytest` inside a source tree without installing the package, I would suggest to put this in a try/except:
```
try:
    from .version import version
except ImportError:  # Can happen in source checkout.
    pass
else:
    packagename = os.path.basename(os.path.dirname(__file__))
    TESTED_VERSIONS[packagename] = version

File: {{ cookiecutter.package_name }}/{{ cookiecutter.module_name }}/conftest.py [hunk 47]
```diff
@@ -3,47 +3,37 @@
...
-## as follow (although default should work for most cases).
-## To ignore some packages that produce deprecation warnings on import
-## (in addition to 'compiler', 'scipy', 'pygments', 'ipykernel', and
-## 'setuptools'), add:
-##     modules_to_ignore_on_import=['module_1', 'module_2']
-## To ignore some specific deprecation warning messages for Python version
-## MAJOR.MINOR or later, add:
-##     warnings_to_ignore_by_pyver={(MAJOR, MINOR): ['Message to ignore']}
-# enable_deprecations_as_exceptions()
+    from pytest_astropy_header.display import PYTEST_HEADER_MODULES, TESTED_VERSIONS
```
mhvk: This probably should also be in a `try/except` - there is no need to insist that packages have this installed. Maybe move it inside the `pytest_configure()`?

{{ cookiecutter.package_name }}/{{ cookiecutter.module_name }}/conftest.py
```diff
@@ -3,47 +3,37 @@
 import os
```
mhvk: Maybe add a comment to the top of the file that this file is meant in particular for running tests from within the interpreter?

pllim:
Oh yeah, @astrofrog , should there be a mention for those to push their releases to `conda-forge` about replicating `pyproject.toml` stuff in the recipe, or is that a separate PR? I think we talked about it on Slack.

bsipocz:
> In general I agree that this is not ideal and really moving conftest.py to the root of the package would be the 'proper' way to do it and then not supporting running self-tests inside installed packages (i.e. the test() command). But I'm pretty sure a number of people feel strongly about keeping that.

Very strongly indeed. But mine are not as strong as @eteq's, so if you convince him I could come on board, too. It proved to be super useful to test user installs, etc. and as I recall was one of the requirement to make work before removing the helpers.

bsipocz:
> One alternative I want to put out there is to simply get rid of the list of dependencies in the pytest header and instead just rely on commands like pip freeze or conda list if people need to check installed packages. In fact that's why I added pip freeze as a command inside the tox config, because that gives a much more complete picture of things. This is my personal preference - to keep the custom header for the other bits but to not actually show the package list.

Please keep the list there. I use it all the time, both locally and CI. It's convenient, and also accurately points to the versions that are actually being used, while pip freeze etc can be tricked with paths etc. to show something else than the actual python interpreter sees.

I'm more flexible with the full list, but as we already talked before, populating it with irrelevant stuff adds noise, while could miss important but assumed dependencies. Comparing builds using the header is also super quick, but it again needs that the list should be the same, and should thus list when something is not installed in one, but it is in the other.

mhvk:
Having done a second package now, which has more of a mix of relevant and less relevant dependencies, my preference would be for the ability to provide a list to be in `setup.cfg` using the `astropy_header_packages` entry (which is already possible and works well), and then have the install generate `package/conftest.py` with the appropriate packages scraped together).

{{ cookiecutter.package_name }}/{{ cookiecutter.module_name }}/conftest.py
astrofrog: Ah good point - in fact we already have a try/except to import the version to the root of the package as ``__version__`` so I just updated the import in the ``conftest.py`` to be ``from . import __version__``.

{{ cookiecutter.package_name }}/{{ cookiecutter.module_name }}/conftest.py
astrofrog: I can't easily move it inside the pytest_configure because the first clause of the 'if' does need to be outside of the function, as it imports ``pytest_*`` functions. However I've adjusted the code so that it should work even if the plugin isn't installed.

{{ cookiecutter.package_name }}/{{ cookiecutter.module_name }}/conftest.py
astrofrog: Just to be clear it also gets picked up when running vanilla pytest, it just doesn't work right when using tox for now. But I've now added a mention that it needs to be in the package for the interpreter version to work.

astrofrog:
@pllim @mhvk @bsipocz - I've pushed a few more changes:

* Fixed edit_on_github to always use master
* Fixed the version specification for the current package in conftest.py
* Made pytest-astropy-header optional
* Added a note about why conftest.py has to live in the source tree for now
* Added a note about updating conda-forge recipes

For the issue of tox not picking up the conftest.py file in time, I'd like to defer this to a follow up PR if possible. I agree that the solution of specifying the packages in ``setup.cfg`` and somehow having that information be copied into the installed version would be nice, but I'd like to investigate that in a dedicated PR. I've opened an issue to keep track of this: https://github.com/astropy/package-template/issues/439 - so can we ignore it for the purposes of the PR and consider it an ongoing bug?

docs/ape17.rst
```diff
@@ -564,3 +564,48 @@ You should also add ``pip-wheel-metadata`` to your ``.gitignore`` file.
...
+in the astropy documentation. The two main changes are that you no longer need
+to manually update the version number in files, instead the version number
+is based on the latest git tag, and in addition the source file should be
+built using the `pep517 <https://pypi.org/project/pep517/>`_ package.
+
+Note on conda recipes and pyproject.toml
+----------------------------------------
+
+While not something you can do until you release your updated package, you will
+need to take care to update conda recipes (e.g. in conda-forge) for your
```
pllim:
```suggestion
need to take care to update conda recipes (e.g., in conda-forge) for your
```

mhvk:
I don't think I'll have time for another detailed look today, so would suggest to just merge this - it was already great as it was and clearly is now even better! And we have follow-up issues raised already.

Super-nice work!

pllim:
Merged. Thank you, all!

Issue #403, opened 2019-05-24
Ability to not have an extern folder?


nabobalis:
While we could turn off the cextern folder, it would be nice to avoid also the normal version as well.

astrofrog:
Agree, in fact most packages shouldn’t have an extern by default.

PR #446, opened 2020-02-07
Remove cextern directory based on input

larrybradley:
Currently the `cextern` directory is always created and thus `include_cextern_folder` has no effect.

fixes https://github.com/astropy/package-template/issues/403


astrofrog:
I wonder whether we even still need this option - 99% of packages won't need this. What do people think about just removing it altogether?

pllim:
FWIW removing it should be a separate PR. This is a simple fix to an existing feature, so I'd vote for accepting it as-is.

PR #449, opened 2020-02-21
Add option to use GitHub Actions for CI

tepickering:
While not perfect, I've been finding Github Actions to be a lot easier and more convenient to use overall than Travis and Azure. Here is a basic configuration that does a matrix of python tests across OS (though only Ubuntu by default), python version, and tox environment. I used this branch to create a test repository for verifying that the Actions workflow does work. Results from a successful test run can be viewed at https://github.com/tepickering/pkgtest/actions/runs/43196684. 

EDIT: Address part of #499


karllark:
Great to see github actions as potentially part of the template.  I used this PR to transition one of my repositories.  Very useful, especially as travis was becoming more and more cranky with time.  Will be doing making the same transition for a bunch of repositories.  Would be a great if this was officially part of the cookiecutter method.

pllim (commented):
@tepickering , I am so glad that you had the foresight to take a stab at this. Are you willing to continue to work on this? Thank you!

* Rebase to remove conflict.
* Remove Travis CI altogether from the template.

{{ cookiecutter.package_name }}/.github/workflows/tox-tests.yml
```diff
@@ -0,0 +1,104 @@
+# GitHub Actions workflow for testing and continuous integration.
+#
+# This file performs testing using tox and tox.ini to define and configure the test environments.
+
+name: Python Tests
+
+on: [push, pull_request]
```
pllim, line 7: Maybe also a good idea to include cron job (say, weekly)?

{{ cookiecutter.package_name }}/.github/workflows/tox-tests.yml
```diff
@@ -0,0 +1,104 @@
...
+
+jobs:
+
+  # Set up matrix to run tox tests across lists of os, python version, and tox environment
+  matrix_tests:
+    runs-on: ${{ matrix.os }}
+    if: "!contains(github.event.head_commit.message, '[ci skip]')"  # github actions doesn't yet support this natively so add here
```
pllim, line 14: This incantation doesn't work anymore. I tried and failed in astropy/astropy#11044

{{ cookiecutter.package_name }}/.github/workflows/tox-tests.yml
```diff
@@ -0,0 +1,104 @@
...
+      matrix:
+        # Github actions supports ubuntu, windows, and macos virtual environments:
+        # https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners
+        #
+        # Only run on ubuntu by default, but can add other os's to the test matrix here.
+        # For example -- os: [ubuntu-latest, macos-latest, windows-latest]
+        os: [ubuntu-latest]
+
+        # Test python 3.7 and 3.8 by default.
+        python-ver: [7, 8]
```
pllim, line 25: To future proof (in case Python 4 comes out), maybe we also specify major version here.
```suggestion
        python-ver: [3.7, 3.8]
```

{{ cookiecutter.package_name }}/.github/workflows/tox-tests.yml
```diff
@@ -0,0 +1,104 @@
...
+      run: |
+        python -m pip install --upgrade pip
+        python -m pip install tox
+    - name: Test with tox
+      run: |
+        tox -e py3${{ matrix.python-ver }}-${{ matrix.tox-env }}
+    #
+    # # This is an example of how to upload coverage to codecov
+    # - name: Upload coverage to codecov
+    #   if: matrix.tox-env == 'cov' && matrix.python-ver == '8'
```
pllim, line 45: @nstarman has an alternative approach at astropy/astropy#11040

Also, I don't think you need a token anymore for Actions.

{{ cookiecutter.package_name }}/.github/workflows/tox-tests.yml
```diff
@@ -0,0 +1,104 @@
...
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.x'
+    - name: Install base dependencies
+      run: |
+        python -m pip install --upgrade pip
+        python -m pip install tox
+    - name: Testing building of docs using tox
+      run: |
+        tox -e build_docs
```
pllim, line 68: I still prefers to use RTD PR build for this, but @Cadair feels the opposite. So 🤷

{{ cookiecutter.package_name }}/.github/workflows/tox-tests.yml
```diff
@@ -0,0 +1,104 @@
...
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.x'
+    - name: Install base dependencies
+      run: |
+        python -m pip install --upgrade pip
+        python -m pip install tox
+    - name: Check links in docs using tox
+      run: |
+        tox -e linkcheck
```
pllim, line 86: From my experience, checking links every push even isn't that useful. I like to put this in cron job (daily or weekly is up to the maintainer).

pllim (commented):
p.s. The way this is set up now, maybe want to consider stages. Otherwise, maybe follow something like what is in astropy core lib to enable "fast fail". (Open to discussions.)

tepickering:
thanks for the review @pllim! now that astropy has taken the actions plunge, circling back to fix this up is high on my list. will get on it asap...

{{ cookiecutter.package_name }}/.github/workflows/tox-tests.yml
tepickering: agreed. if nothing else, then as an example how to do it.

{{ cookiecutter.package_name }}/.github/workflows/tox-tests.yml
tepickering: oh bother. looks like there's a new option on the marketplace that might do the trick: https://github.com/marketplace/actions/ci-skip-action. it's 3rd party, though. the way forward for astropy as a whole for now may be to fork that and modify as needed for our purposes. for now i'll remove these from the template until we have a workable solution.

{{ cookiecutter.package_name }}/.github/workflows/tox-tests.yml
tepickering: yeah, i got bit by this already for different reasons so was planning to make this change, anyway.

{{ cookiecutter.package_name }}/.github/workflows/tox-tests.yml
tepickering: coverage handling also needs fixing up to merge coverage output from multiple tox environments. this hack specifically uploads from one specific environment while the one in astropy uploads from each environment containing `-cov` so, afaict, whatever ends up on codecov at the end is whatever the last one uploaded happened to be. i have bookmarked a recipe for how to combine coverage from multiple tox envs so i'll dig that out and see if i can apply it.

and, indeed, no codecov token is required any longer for public repos.

{{ cookiecutter.package_name }}/.github/workflows/tox-tests.yml
tepickering: i'm agnostic. RTD is the real test since that's where they're published, ultimately. otoh, if actions is used to determine whether something is ok to merge, it's useful to make successful doc building a part of that process. in other words, por que no los dos? (insert why_not_both.gif)

{{ cookiecutter.package_name }}/.github/workflows/tox-tests.yml
tepickering: good point. when editing docs, i've found the link check test useful for catching typos or bad redirects. however, that can be done locally easy enough so a scheduled test for CI is probably best for most usage cases.

astrofrog:
I would vote with not having stages which add complication

pllim:
Thanks for your patience, @tepickering ! I will be happy to have another look after the next round of changes, so feel free to ping me. 😸

{{ cookiecutter.package_name }}/.github/workflows/tox-tests.yml
saimn: I also prefer the full version number, and the more explicit `python-version` which we are using in the other repos.

saimn: Also it would be good to add an example of a "manual" matrix, with `include:`. Because it may not be obvious for people that the current example will run all combinations (which will result in a huge number of jobs if they add some python versions and OS).

{{ cookiecutter.package_name }}/.github/workflows/tox-tests.yml
```diff
@@ -0,0 +1,104 @@
...
+        # For example -- os: [ubuntu-latest, macos-latest, windows-latest]
+        os: [ubuntu-latest]
+
+        # Test python 3.7 and 3.8 by default.
+        python-ver: [7, 8]
+
+        # Specify which tox environments to test in this list.
+        tox-env: [cov, alldeps, devdeps, astropylts]
+    steps:
+    - uses: actions/checkout@v2
```
saimn, line 30: If the template also use setuptools-scm (?), this should be a full checkout.

tepickering: i changed this to do `fetch_depth: 0` to match what astropy is doing. not a huge deal for small/new repos. helped a lot for some of my older, cruftier ones...

{{ cookiecutter.package_name }}/.github/workflows/tox-tests.yml
tepickering: i ended up punting on the complicated path in favor of what astropy and photutils do: just have one test run with coverage enabled and upload those results.

tepickering:
i renamed the main workflow file and significantly cleaned it up to more closely match what astropy and photutils are doing in terms of structure. namely, i did away with the multi-level matrix in favor of bespoke steps to explicitly choose what to do, where, and how.

tepickering:
did some more expunging of references to travis and updated the docs to reflect the changes. i see that the template itself is still using travis for its own CI. migrating that is probably best left to a different PR though...

pllim:
> the template itself is still using travis for its own CI

Yes, @Cadair said he would deal with that separately, I think.

TEMPLATE_CHANGES.md
```diff
@@ -13,11 +13,13 @@ be copied over manually if desired.
 - Moved as much as possible of the setup.py configuration to setup.cfg, and now require
   setuptools>=30.3. [#375]
 
-- Removed Appveyor CI option. Use ``os: windows`` in Travis CI. [#420]
+- Removed Appveyor CI option. Use ``os: windows-latest`` in Github Actions. [#420]
 
 - Refactored the template to follow the recommendations in APE 17:
   https://github.com/astropy/astropy-APEs/blob/master/APE17.rst [#438]
 
+- Switch to using Github Actions for CI; remove Travis config.
```
pllim, line 21:
```suggestion
- Switch to using Github Actions for CI; remove Travis config. [#449]
```

cookiecutter.json
```diff
@@ -10,11 +10,14 @@
...
     "github_project": "astropy/astropy",
-    "use_travis_ci": "y",
+    "use_gh_actions": "y",
     "use_read_the_docs": "y",
     "sphinx_theme": "astropy-bootstrap",
     "_parent_project": "astropy",
     "required_dependencies": "astropy",
     "optional_dependencies": "",
-    "minimum_python_version": ["3.6", "3.7", "3.8"]
+    "minimum_python_version": ["3.6", "3.7", "3.8"],
```
pllim, line 19: I wonder if we also should bump minversion options, but I know it is out of scope, so it is okay to ignore this one.
```suggestion
    "minimum_python_version": ["3.7", "3.8", "3.9"],
```

docs/ape17.rst
```diff
@@ -496,11 +496,10 @@ recommended approach is to use the tox file to set up the different
 configurations you want to use, and to then keep the CI configuration as simple
 as possible.
 
-If you use Travis CI, a good place to start is the ``.travis.yml`` file
-generated in :ref:`Step 0 <step-rerender>`, and you can then see if any previous customizations you had
-made need to be copied over. This file shows how one can configure Travis to use
-tox, optionally using conda via ci-helpers to set up Python on MacOS X and
-Windows.
+Github now has an integrated CI service, Github Actions. If you wish to use Actions, a good place to start
```
pllim, line 499: Officially spelling is "GitHub" with a capital "H" so I think we need to follow that throughout for consistency?
```suggestion
GitHub now has an integrated CI service, GitHub Actions. If you wish to use Actions, a good place to start
```

docs/nextsteps.rst
```diff
@@ -15,37 +15,32 @@ Setting Up Continuous Integration
...
-``.travis.yml``. Generally you should aim to always have your ``master`` branch
+services continuously test your package
+for each commit. Every pull request against your main repository will be
+automatically tested and failing tests will be flagged by these services.
+
+Github Actions
+##############
+
+Github now provides an integrated CI service called `Github Actions <https://docs.github.com/en/free-pro-team@latest/actions>`__.
+The default workflows in ``.github/workflows`` show how to set up perform integration testing
```
pllim, line 26:
```suggestion
The default workflows in ``.github/workflows`` show how to set up integration testing
```

docs/nextsteps.rst
```diff
@@ -15,37 +15,32 @@ Setting Up Continuous Integration
...
+upon every push or pull request, ``ci_tests.yml``, and how to run scheduled tests via cron, ``ci_cron_weekly.yml``.
+The default ``ci_tests.yml`` file contains a large number of builds against various versions of Python, astropy, and
+numpy, and you should choose the ones relevant to your project. Generally you should aim to always have your ``master`` branch
```
pllim, line 29:
```suggestion
numpy, and you should choose the ones relevant to your project. Generally, you should aim to always have your ``master`` branch
```

hooks/post_gen_project.py
```diff
@@ -39,8 +39,8 @@ def process_license(license_name):
 
     process_license('{{ cookiecutter.license }}')
 
-    if '{{ cookiecutter.use_travis_ci }}' != 'y':
-        remove_file('.travis.yml')
+    if '{{ cookiecutter.use_gh_actions }}' != 'y':
+        remove_dir('.github')
```
pllim, line 43: Just to note that in the future if the template also wants to include GitHub issue/PR templates, this line needs to be revisited to be more specific to just remove the Actions files.

{{ cookiecutter.package_name }}/.github/workflows/ci_tests.yml
```diff
@@ -0,0 +1,79 @@
+# GitHub Actions workflow for testing and continuous integration.
+#
+# This file performs testing using tox and tox.ini to define and configure the test environments.
+
+name: CI Tests
+
+on: [push, pull_request]
```
pllim, line 7: Should we also include @saimn 's patch in astropy/astropy#11054 so Actions won't also run on forks when PR is opened?

pllim (commented):
Looking great. Thanks! Just minor comments.

hooks/post_gen_project.py
tepickering: ah, good point. i'll fix that.

{{ cookiecutter.package_name }}/.github/workflows/tox-tests.yml
tepickering: yeah, i switched out my matrix for a "manual" that's more keeping in the style of what astropy and other affiliated packages are doing.

cookiecutter.json
tepickering: i think we might as well do this now even if the actions test matrix only covers 3.7 and 3.8 for now.

docs/ape17.rst
tepickering: yeah, i see it done both ways, but the official docs use the H so we should, too.

{{ cookiecutter.package_name }}/.github/workflows/ci_tests.yml
tepickering: yes, this keeps the template in-line with what astropy is doing so definitely worth doing.

nstarman:
The old travis setup split the tests into sections: "initial" and "comprehensive". I recently switched to GitHub CI and have found this to be possible by splitting the tests into jobs, setting ``fail-fast=False`` within each job and setting up job dependencies.

As an example,

[ci_tests.txt](https://github.com/astropy/package-template/files/5555843/ci_tests.txt)

{{ cookiecutter.package_name }}/.github/workflows/ci_tests.yml
```diff
@@ -0,0 +1,84 @@
...
+jobs:
+  # Github Actions supports ubuntu, windows, and macos virtual environments:
+  # https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners
+  ci_tests:
+    name: ${{ matrix.name }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        include:
+          - name: Code style checks
```
nstarman, line 23: If the code style check fails then no other test is run. I implement a fix for this in the example file I included in https://github.com/astropy/package-template/pull/449#issuecomment-729160415

pllim:
> sections: "initial" and "comprehensive"

@astrofrog already stated his objection to re-introducing stages in Actions due to increased complexity.

nstarman:
> increased complexity.

Is there any other means to prevent a code linter check from stopping the other tests?

{{ cookiecutter.package_name }}/.github/workflows/ci_tests.yml
```diff
@@ -0,0 +1,84 @@
...
+    - master
+    tags:
+  pull_request:
+
+jobs:
+  # Github Actions supports ubuntu, windows, and macos virtual environments:
+  # https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners
+  ci_tests:
+    name: ${{ matrix.name }}
+    runs-on: ${{ matrix.os }}
```
nstarman, line 19: Someone kindly gave me this piece of code to allow for skipping CI

``if: "!(contains(github.event.head_commit.message, '[skip ci]') || contains(github.event.head_commit.message, '[ci skip]'))"``

{{ cookiecutter.package_name }}/.github/workflows/ci_tests.yml
```diff
@@ -0,0 +1,84 @@
...
+    - name: Install base dependencies
+      run: |
+        python -m pip install --upgrade pip
+        python -m pip install tox codecov
+    - name: Test with tox
+      run: |
+        tox -e ${{ matrix.toxenv }}
+    # This is an example of how to upload coverage to codecov
+    # - name: Upload coverage to codecov
+    #   if: "endsWith(matrix.toxenv, '-cov')"
```
nstarman, line 81: ``contains(matrix.toxenv,'-cov')`` might be better

{{ cookiecutter.package_name }}/.github/workflows/ci_tests.yml
```diff
@@ -0,0 +1,84 @@
...
+# This file performs testing using tox and tox.ini to define and configure the test environments.
+
+name: CI Tests
+
+on:
+  push:
+    branches:
+    - master
+    tags:
+  pull_request:
```
nstarman, line 12: Might be too much of an edge case, but adding
```
  pull_request:
    branches:  # only build on PRs against main
      - master
```

likewise limits PR builds. I have this set on my private repos to preserve minutes (public too if I forgot to comment it out).

pllim: Probably too specific to be useful. Most repos don't do branch and for those that do, they probably want to test even if a PR is opened against older release branch (especially if they are following gitflow).

pllim:
> Is there any other means to prevent a code linter check from stopping the other tests?

Without stages and with "fail fast" on (the default), you hope that it finishes first, or you make it into another job. But I think in some cases, you actually want a failing linter check to stop other jobs. 🤷

{{ cookiecutter.package_name }}/.github/workflows/ci_tests.yml
pllim: Does that still work? It didn't work for me when I tried at #11044 . In fact, I dumped the whole `github.event` context out and couldn't find any `head_commit` in it.

nstarman: I haven't tested it recently. It worked when I first added it and doesn't NOT work. But I see 11044 is more recent than my last test, so I guess it doesn't work anymore 👎 .

{{ cookiecutter.package_name }}/.github/workflows/ci_tests.yml
tepickering: agreed.

tepickering:
i think in the use case of starting from scratch with a fresh code base, having the linter check fail fast and stop everything else is useful. shimming an actions workflow into an existing codebase is a different use case where you either want to allow the check to fail somehow or adjust the check to only fail in the worst cases.

{{ cookiecutter.package_name }}/.github/workflows/ci_tests.yml
tepickering: this actually reminds me that we should be changing `master` here to `main` which is GitHub's new default...

{{ cookiecutter.package_name }}/.github/workflows/ci_tests.yml
tepickering: i had this in the template originally, but took it out because it's no longer a reliable method of doing this. i would prefer to leave this alone until either it's supported natively (https://github.com/actions/runner/issues/774) or there's an official action published that will do this properly. a 3rd party action exists, https://github.com/marketplace/actions/ci-skip-action, and seems to work well. astropy may wish to make their own version of this if native support is not forthcoming.

larrybradley: I'm pretty sure this worked for me in `photutils` when I tested it:
https://github.com/astropy/photutils/blob/master/.github/workflows/ci_tests.yml#L18

{{ cookiecutter.package_name }}/.github/workflows/ci_tests.yml
tepickering: as i said below, i think having code style checks fail and stop everything is appropriate when starting with a fresh code base. again, this is a use case where better upstream support is required: https://github.com/actions/toolkit/issues/399. both your workaround and using `continue-on-error` cause the code style check to still show up as a failure that would still block merging. i would advocate against implementing those workarounds here and would rather wait for proper `allow-failure` support.

{{ cookiecutter.package_name }}/.github/workflows/ci_tests.yml
pllim: Nope, no work astropy/photutils#1121

{{ cookiecutter.package_name }}/.github/workflows/tox-tests.yml
tepickering: i did move the link check to a cron scheduled workflow as an example of how to do so.

tepickering:
i built a dummy repo from this branch and you can see the results of the CI here: 

https://github.com/tepickering/tep_test/actions/runs/369174337

a bit of light whingeing about `ubuntu-latest` moving to `ubuntu-20.04` soon, but clean otherwise.

QUESTIONS.rst
```diff
@@ -16,7 +16,7 @@ a description of each of the prompts.
 9. ``use_compiled_extensions``: Whether you plan to use compiled extensions in your package
 10. ``edit_on_github_extension``: Set to ``True`` to enable the edit on GitHub sphinx extension.
 11. ``github_project``: This is the GitHub identifier for the edit on GitHub extension and the changelog link extension.
-12. ``use_travis_ci``: If ``'y'`` the template will include an example ``.travis.yml`` file for the Travis CI service.
+12. ``use_gh_actions``: If ``'y'`` the template will include an example ``.github/workflows/tox-tests.yml`` file for the Github Actions CI service.
```
pllim, line 19:
```suggestion
12. ``use_gh_actions``: If ``'y'`` the template will include an example ``.github/workflows/tox-tests.yml`` file for the GitHub Actions CI service.
```

TEMPLATE_CHANGES.md
```diff
@@ -13,11 +13,13 @@ be copied over manually if desired.
 - Moved as much as possible of the setup.py configuration to setup.cfg, and now require
   setuptools>=30.3. [#375]
 
-- Removed Appveyor CI option. Use ``os: windows`` in Travis CI. [#420]
+- Removed Appveyor CI option. Use ``os: windows-latest`` in Github Actions. [#420]
```
pllim, line 16:
```suggestion
- Removed Appveyor CI option. Use ``os: windows-latest`` in GitHub Actions. [#420]
```

TEMPLATE_CHANGES.md
```diff
@@ -13,11 +13,13 @@ be copied over manually if desired.
...
 
 - Refactored the template to follow the recommendations in APE 17:
   https://github.com/astropy/astropy-APEs/blob/master/APE17.rst [#438]
 
+- Switch to using Github Actions for CI; remove Travis config. [#449]
```
pllim, line 21:
```suggestion
- Switch to using GitHub Actions for CI; remove Travis config. [#449]
```

docs/ape17.rst
```diff
@@ -496,11 +496,10 @@ recommended approach is to use the tox file to set up the different
...
 
-If you use Travis CI, a good place to start is the ``.travis.yml`` file
-generated in :ref:`Step 0 <step-rerender>`, and you can then see if any previous customizations you had
-made need to be copied over. This file shows how one can configure Travis to use
-tox, optionally using conda via ci-helpers to set up Python on MacOS X and
-Windows.
+GitHub now has an integrated CI service, GitHub Actions. If you wish to use Actions, a good place to start
+is the ``.github/workflows/ci_tests.yml`` file generated in :ref:`Step 0 <step-rerender>`. You can then
+see if any previous customizations you had made need to be copied over. This file shows how one can configure
+Actions to use tox to test different environments with different versions of python on different platforms.
```
pllim, line 502:
```suggestion
Actions to use tox to test different environments with different versions of Python on different platforms.
```

docs/nextsteps.rst
```diff
@@ -15,37 +15,32 @@ Setting Up Continuous Integration
...
-
-There  are many `example build configuations
-<https://github.com/astropy/astropy/wiki/travis-ci-test-status>`__ for
-``.travis.yml``. Generally you should aim to always have your ``master`` branch
-work with the latest stable and latest development version of astropy (i.e. the
+services continuously test your package
+for each commit. Every pull request against your main repository will be
+automatically tested and failing tests will be flagged by these services.
+
+Github Actions
```
pllim, line 22:
```suggestion
GitHub Actions
```

docs/nextsteps.rst
```diff
@@ -15,37 +15,32 @@ Setting Up Continuous Integration
...
+##############
+
+Github now provides an integrated CI service called `Github Actions <https://docs.github.com/en/free-pro-team@latest/actions>`__.
```
pllim, line 25:
```suggestion
GitHub now provides an integrated CI service called `GitHub Actions <https://docs.github.com/en/free-pro-team@latest/actions>`__.
```

docs/nextsteps.rst
```diff
@@ -15,37 +15,32 @@ Setting Up Continuous Integration
...
+The default workflows in ``.github/workflows`` show how to set up integration testing
+upon every push or pull request, ``ci_tests.yml``, and how to run scheduled tests via cron, ``ci_cron_weekly.yml``.
+The default ``ci_tests.yml`` file contains a large number of builds against various versions of Python, astropy, and
+numpy, and you should choose the ones relevant to your project. Generally, you should aim to always have your
+``main`` branch work with the latest stable and latest development version of astropy (i.e. the
```
pllim, line 30:
```suggestion
``main`` branch work with the latest stable and latest development version of astropy (i.e., the
```

docs/options.rst
```diff
@@ -18,7 +18,7 @@ a description of each of the prompts:
 #. ``use_compiled_extensions``: Whether you plan to use compiled extensions in your package
 #. ``edit_on_github_extension``: Set to ``True`` to enable the edit on GitHub sphinx extension.
 #. ``github_project``: This is the GitHub identifier for the edit on GitHub extension and the changelog link extension.
-#. ``use_travis_ci``: If ``'y'`` the template will include an example ``.travis.yml`` file for the Travis CI service.
+#. ``use_gh_actions``: If ``'y'`` the template will include an example ``.github/workflows/tox-tests.yml`` file for the Github Actions CI service.
```
pllim, line 21:
```suggestion
#. ``use_gh_actions``: If ``'y'`` the template will include an example ``.github/workflows/tox-tests.yml`` file for the GitHub Actions CI service.
```

pllim (approved):
Just some wording nitpicks, otherwise LGTM. Thank you so much!

QUESTIONS.rst
tepickering: oops, missed that.

pllim:
Thanks, @tepickering and everyone who reviewed!

PR #470, opened 2020-05-24
Remove warning about APE 17 migration guide

astrofrog:
Might be worth waiting a little in case anyone else replies on the core maintainers mailing list, but I think the warning is ready to be removed from the migration guide.

Once this is removed, we should probably start advertising the deprecation of astropy-helpers and the existence of the migration guide on astropy@python.org and astropy-dev.

### DIFF
--- a/docs/ape17.rst
+++ b/docs/ape17.rst
@@ -1,9 +1,6 @@
 APE 17 Migration Guide
 ======================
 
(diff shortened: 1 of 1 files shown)


Cadair:
:shipit: ?

astrofrog:
Let's do this!

pllim:
download

PR #481, opened 2020-10-04
private_project cookiecutter question

nstarman:
On private repositories, every PR is build twice by the CI: once for the branch and once for the PR.
This changes it so that PRs are only built once, but still allows the master branch to be built after a PR.
Other branches are not built unless in a PR.

This option makes it easier to do code-dev on a project before taking it public.
I have added instructional docs about how to un-implement the private-repo options and take a project public.

In future, when further private-project options are added, they can be similarly documented.

### DIFF
--- a/rendered.yml
+++ b/rendered.yml
@@ -17,3 +17,4 @@ default_context:
   minimum_python_version: "3.6"
   required_dependencies: "astropy"
   optional_dependencies: ""
+  private_project: "n"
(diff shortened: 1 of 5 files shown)


nstarman:
I need to add this to the changelog

rendered.yml
```diff
@@ -17,3 +17,4 @@ default_context:
   minimum_python_version: "3.6"
   required_dependencies: "astropy"
   optional_dependencies: ""
+  private_project: "y"
```
Cadair, line 20: Would this not be better to render the default as no?

```suggestion
  private_project: "n"
```

nstarman: Agreed. I think that was for a test I did.

nstarman:
@Cadair , squashed and passing.

nstarman:
Now that we don't use travis for CI, this PR request doesn't do anything.

PR #483, opened 2020-10-04
Conftest configure

nstarman:
- the conftest code is given a doctoring from the top-level comments
- the `pytest_configure` function is likewise documented
- `import astropy.units as u` is auto-imported into docstring tests

### DIFF
--- a/{{ cookiecutter.package_name }}/{{ cookiecutter.module_name }}/conftest.py
+++ b/{{ cookiecutter.package_name }}/{{ cookiecutter.module_name }}/conftest.py
@@ -1,7 +1,11 @@
-# This file is used to configure the behavior of pytest when using the Astropy
-# test infrastructure. It needs to live inside the package in order for it to
-# get picked up when running the tests inside an interpreter using
-# packagename.test
+"""Configure Test Suite.
+
+This file is used to configure the behavior of pytest when using the Astropy
+test infrastructure. It needs to live inside the package in order for it to
+get picked up when running the tests inside an interpreter using
+packagename.test
+
+"""
 
 import os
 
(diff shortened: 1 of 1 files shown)


nstarman:
While I agree with Zen of Python line 2, I feel this is a somewhat special case. On the developer side, anyone writing example code in docstrings either sticks this in boilerplate code at the beginning of the section or, if it's critical, purposefully includes it with the specific example. While this doesn't change the latter, it nicely eliminates the former. Examples are just the core code.
On the user side, this feels like a standard import that astropy users just know, like `numpy -> np` (which actually might be a good fixture to add).

embray:
I agree with @pllim : It's cool to see that this is *possible* (I did not know this) but I think it's too magical, implicit, confusing (a new user stumbling on a particular doc does not necessarily know where `np` or `u` came from).  It's also arbitrary bc I can think of several other "common imports" I would normally like to skip having to write when writing tests.  At the end of the day, using implicit magic is also too prone to hide mistakes.

nstarman:
Okay. Outvoted 🙃. Not that this is a democracy. I'll drop the pytest auto-injections.

nstarman:
#no-changelog-needed

astrofrog (commented):
Thanks for the PR! Just out of curiosity what is the motivation for adding docstrings to the module and function? This file shouldn't appear in public API docs in general - pytest_configure for example is not something users should ever call directly, and the module should ideally be excluded from at API docs.

nstarman:
It's mostly personal reference. Even for short code snippets, with python being such a dynamical language I like to know what I (or someone else) considered the scope of the function.

embray:
I don't see a problem with it--actually I think it makes sense.   I think all modules should be documented with docstrings instead of comments, just for consistency's sake, even if it's not a docstring most people will ever see at runtime.

embray:
This isn't doing anything more than adding the docstrings now, but I don't see the harm either.  Thanks @nstarman !  I'll try to follow up on some of your other PRs sooner rather than later.

PR #488, opened 2020-10-05
Add pandoc for jupyter

nstarman:
jupyter notebooks require pandoc to show in docs. This is added to travis.


nstarman:
Given discussion in https://github.com/astropy/package-template/pull/482#discussion_r499767226, I think this PR should be closed.

PR #495, opened 2020-10-17
modindex_common_prefix

nstarman:
This simplifies the search page.

From https://www.sphinx-doc.org/en/master/usage/configuration.html

> A list of prefixes that are ignored for sorting the Python module index (e.g., if this is set to ['foo.'], then foo.bar is shown under B, not F). This can be handy if you document a project that consists of a single package. Works only for the HTML builder currently. Default is [].
>
> New in version 0.6.


nstarman:
@pllim , The travis PR actually successfully built but is failing to report back. This PR is ready for review.

pllim (approved):
Yeah, I like this. Definitely helps with the listing. Thanks!

PR #506, opened 2021-03-09
MNT: Replace master with main

pllim:
This PR attempts to replace all mention of `master` to `main` in this repository. This should be reviewed and merged right after the maintainer has renamed the default branch.

This is an automated update made by the `batchpr` tool :robot: - feel free to close if it doesn't look good! You can report issues to @pllim.

This is part of #505

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -10,7 +10,7 @@ on:
 # https://github.com/actions/download-artifact/issues/60
 jobs:
   upload:
-    name: Push template to master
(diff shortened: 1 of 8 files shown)


pllim:
Depends on other parts of the Project a little...

pllim:
Some of the doc needs immediate update because upstream packages have changed default branch. So, we can either push ahead with the `master -> main` transition too, or if you want, I can take out the subset of the changes that are already relevant into a separate PR.

PR #510, opened 2021-04-13
Remove duplicate entries in .gitignore

larrybradley:
These entries are also listed under the "Eclipse editor project files" section.


pllim (approved):
Seems uncontroversial enough. Good catch!

PR #511, opened 2021-04-14
TST: Use nightly astropy

### LINKED ISSUE #468: Devdeps builds take forever. Any suggestions?
Hi, I noticed that devdeps builds take a lot of time to run, because the tox initialization takes forever. I tried specifying the #egg option for every development dependency and, even if it does speed up things a lot if I call pip directly, it does not seem to affect the tox build.
Any ideas?

pllim:
Fix #468 

Also see astropy/astropy#11522

(Nothing much we can do about the other stuff like `stingray` but we got `numpy` and `astropy` covered already.)

### DIFF
--- a/{{ cookiecutter.package_name }}/tox.ini
+++ b/{{ cookiecutter.package_name }}/tox.ini
@@ -12,6 +12,7 @@ requires =
 isolated_build = true
 indexserver =
     NIGHTLY = https://pypi.anaconda.org/scipy-wheels-nightly/simple
+    NIGHTLY2 = https://pkgs.dev.azure.com/astropy-project/astropy/_packaging/nightly/pypi/simple
 
 [testenv]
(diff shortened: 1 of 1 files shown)


{{ cookiecutter.package_name }}/tox.ini
```diff
@@ -12,6 +12,7 @@ requires =
 isolated_build = true
 indexserver =
```
Cadair, line 13: This indexserver option [is deprecated](https://tox.readthedocs.io/en/latest/config.html#conf-indexserver) so we probably shouldn't be using it here.

pllim: Errr.... How do we set multiple env var to grab numpy-dev and astropy-dev separately from everything else?

PR #516, opened 2021-10-05
Fix typo

lpsinger:


astrofrog:
I don't think this is a typo - it is referring to the test function. But maybe we should add () after test?

lpsinger:
Ohhh... I see.

lpsinger:
Take a look at this version.

{{ cookiecutter.package_name }}/{{ cookiecutter.module_name }}/conftest.py
```diff
@@ -3,7 +3,7 @@
 This file is used to configure the behavior of pytest when using the Astropy
 test infrastructure. It needs to live inside the package in order for it to
 get picked up when running the tests inside an interpreter using
-packagename.test
+`{{ cookiecutter.package_name }}.test()`.
```
astrofrog, line 6: I think it should be module name instead of package name?

astrofrog (changes requested):

{{ cookiecutter.package_name }}/{{ cookiecutter.module_name }}/conftest.py
lpsinger: Fixed.

astrofrog (approved):

PR #521, opened 2022-05-11
Set minimum Python version for readthedocs

lpsinger:


{{ cookiecutter.package_name }}/.readthedocs.yml
```diff
@@ -4,7 +4,7 @@ build:
   image: latest
 
 python:
-  version: 3.7
+  version: '{{ cookiecutter.minimum_python_version }}'
```
pllim, line 7: This won't work. RTD has changed the way they specify Python version for 3.8 or later.

https://github.com/astropy/astropy/blob/main/.readthedocs.yml

lpsinger: Good catch! Please take another look.

{{ cookiecutter.package_name }}/.readthedocs.yml
```diff
@@ -2,9 +2,10 @@ version: 2
 
 build:
   image: latest
```
pllim, line 4:
```suggestion
  os: ubuntu-20.04
```

pllim (approved):
Thanks!

PR #528, opened 2023-05-18
fix: wheel is not needed

henryiii:
This has never been needed when using PEP 517.


pllim (approved):
Thanks!

We deprecated this template (https://github.com/astropy/package-template#deprecation-warning) but this seems uncontroversial enough for one last merge.

